PR #5, opened 2024-01-28
Fix all GMloader scripts with new standards 

Cebion:
from https://portmaster.games/packaging.html

-  Added [ -f "/etc/os-release" ] && source "/etc/os-release" 
- Added new logging method "exec > >(tee "$GAMEDIR/log.txt") 2>&1" 
- Encapsulate Pipewirefix for Jelos in IF statements Fix false LD_LIBRARY_PATH that exluded JELOS
- Added file check for  data.win & game.win mv
- Changed Names of Chip n Dale, thecurseofissyos and others


kloptops (approved):
👍

tabreturn:
Will Portmaster change the names of Chip n Dale, thecurseofissyos and others to match your renaming?  For example, if somebody updates/reinstalles those via the PortMaster GUI, does it install a duplicate game (with different file names), or rename and overwrite the existing files?

kloptops:
> Will Portmaster change the names of Chip n Dale, thecurseofissyos and others to match your renaming? For example, if somebody updates/reinstalles those via the PortMaster GUI, does it install a duplicate game (with different file names), or rename and overwrite the existing files?

No. It only applies the new naming scheme to ports added after the 26th of January.

tabreturn (approved):

PR #9, opened 2024-01-28
Update Descent port

JeodC:
# Update for Descent

## Changes Made
- Change `lib` folder to `libs`
- Reflect change in sh file
- Synchronize with newer PortMaster update from old repo
- Add soundfont, remove .dxa file (user preference
- Add libmodplug.so, which JelOS doesn't provide by default in some builds
- Remove log.txt generation since game generates gamelog.txt in `conf` folder
- Update readme


Cebion (approved):
Approved after review

Tekkenfede (approved):
Approved

PR #24, opened 2024-02-01
Add issue templates

JeodC:
Adds two issue templates, one for the PortMaster GUI Tool and the other for specific Ports.


Cebion (changes requested):
Maybe Replace RG353 with RG351MP.
Also add theRA firmware.

Device support list is vastly incomplete atm.

kloptops (changes requested):
Could the gui template have a section requesting the pugwash.txt file.

kloptops (approved):
That’ll do. 😁

Cebion (approved):

PR #39, opened 2024-02-04
Update solarus engine for existing ports

JeodC:
Provided by Cebion.

All solarus-run and solarus lib files are rebuilt to exclude sdl controls, which removes the need for lua patching and avoids conflicts due to hardcoded gamepad controls.


Cebion (approved):
Tested, works perfect.

kloptops (approved):
Looks good. Might be a good idea to move solarus into a runtime so in the future the ports will just use the latest version. It’ll also help with new platforms being added.

PR #68, opened 2024-02-07
Update to: F-Zero Pocket, Mimi's Delivery Dash, Spelunky

JeodC:
## F-Zero Pocket
- Modify sh to use new modular cfw format

## Mimi's Delivery Dash
- Modify sh to use new modular cfw format
- Add libcrypto

## Spelunky
- Modify sh to use new modular cfw format
- Add gptokeyb controls (fixes broken controls in jelos, maybe elsewhere too)


Cebion (approved):
Looks good, as long as the control fiel is in portmaster_release (non beta)

kloptops (approved):
👍

PR #83, opened 2024-02-12
Remove incorrect note for DevilutionX

AJenbo:
This information is not correct. The GOG data file is identical to the CD release, all releases will work just fine (1.00 CD, 1.04 CD, 1.08 CD, Mac CD, GOG release, BattleNet release). patch_rt.mpq isn't even loaded so having it makes no difference.

Also there are only 16 levels in Diablo, so how would you get stuck on level 20...

Source: I'm an upstream maintainer.

### DIFF
--- a/ports/devilutionx/README.md
+++ b/ports/devilutionx/README.md
@@ -4,8 +4,6 @@ Copy diabdat.mpq from your CD or GoG installation (or extract it from the GoG in
 
 Do not delete the gamecontrollerdb.txt file in the /roms/ports/devilution folder or there will be no controller support in the game! For controls, see here
 
-**Important Note: ** It’s been reported that you must make sure you use the GOG version of diabdat.mpq with the newest patch_rt.mpq or you may experience a freeze of the game around level 20.
-
 If you experience issues after a version upgrade, backup your save file and uninstall/reinstall the port.
 
 If you make a config change that crashes the game, you can delete the ini file to get the game back to a good state


Cebion:
> This information is not correct. The GOG data file is identical to the CD release, all releases will work just fine (1.00 CD, 1.04 CD, 1.08 CD, Mac CD, GOG release, BattleNet release). patch_rt.mpq isn't even loaded so having it makes no difference.
> 
> Also there are only 16 levels in Diablo, so how would you get stuck on level 20...
> 
> Source: I'm an upstream maintainer.

I think the original author meant character level 20.
Saw that here 
https://www.reddit.com/r/Diablo1/comments/pg014y/level_cap_increasing

Cebion (approved):
👍

AJenbo:
That page doesn't appear to mention 20, a crash or the patch mpq.

More likely it's a reference to the Hellfire crypt, we have see reports of the game crashing there where it turned our they where using it pirate copy with a damaged hellfire.mpq

kloptops (approved):
Sounds good to me. 👍

Cebion:
> That page doesn't appear to mention 20, a crash or the patch mpq.
> 
> More likely it's a reference to the Hellfire crypt, we have see reports of the game crashing there where it turned our they where using it pirate copy with a damaged hellfire.mpq

Copy paste error =)
I meant this one https://www.reddit.com/r/RG350/comments/krwncx/devilutionx_freezing_around_level_20/

AJenbo:
Ah ok. Well there is nothing special happening at that hero level, it could be that they start finding items which triggered broken data from what ever file they used. But the post does appear to be dealing with some computer goblins.

There have also been instances where the MPQ file didn't transfer correctly from the CD or to the device over USB, here the solution is usually to try again. If someone doesn't already own Diablo I would recommend the GOG one as it works best on PC, but in combination with DevilutionX there isn't any difference between releases, so long as your files aren't damaged :)

PR #92, opened 2024-02-14
Updated Quakespasm scripts, change controls to address drift, removed libs that interfered with ogg playback on JelOS

Slayer366:
Updated Quakespasm scripts, change controls to address drift, removed libs that interfered with ogg playback on JelOS


Cebion (approved):
:+1:

kloptops (approved):
👍

PR #94, opened 2024-02-14
Update Hexen 2, condense libs, fix ogg playback on JelOS

Slayer366:
Update Hexen 2 scripts, detect devices with triggers, condense libs to one directory, removed libs that interfered with ogg playback on JelOS


Cebion:
@Slayer366 I do have a question.
You're doing this
$ESUDO rm -f $GAMEDIR/lib/libasound.so.2 $GAMEDIR/lib/libvorbis.so.0 $GAMEDIR/lib/libvorbisenc.so.2 $GAMEDIR/lib/libvorbisfile.so.3

Why are you deleting the files in the .sh instead of deleting them from the package?
That does not make sense, only for existing installations maybe.

kloptops:
> @Slayer366 I do have a question. You're doing this $ESUDO rm -f $GAMEDIR/lib/libasound.so.2 $GAMEDIR/lib/libvorbis.so.0 $GAMEDIR/lib/libvorbisenc.so.2 $GAMEDIR/lib/libvorbisfile.so.3
> 
> Why are you deleting the files in the .sh instead of deleting them from the package? That does not make sense, only for existing installations maybe.

Looks like you answered your own question.

Cebion:
> > @Slayer366 I do have a question. You're doing this $ESUDO rm -f $GAMEDIR/lib/libasound.so.2 $GAMEDIR/lib/libvorbis.so.0 $GAMEDIR/lib/libvorbisenc.so.2 $GAMEDIR/lib/libvorbisfile.so.3
> > Why are you deleting the files in the .sh instead of deleting them from the package? That does not make sense, only for existing installations maybe.
> 
> Looks like you answered your own question.

Yes, but then d

> > @Slayer366 I do have a question. You're doing this $ESUDO rm -f $GAMEDIR/lib/libasound.so.2 $GAMEDIR/lib/libvorbis.so.0 $GAMEDIR/lib/libvorbisenc.so.2 $GAMEDIR/lib/libvorbisfile.so.3
> > Why are you deleting the files in the .sh instead of deleting them from the package? That does not make sense, only for existing installations maybe.
> 
> Looks like you answered your own question.

I know, but! after they were removed once, the script will try to delete these files every time.
So either put them in an -f file condition or don't put the delete line in at all.

Slayer366:
I've certainly removed the files from the package, but like you pointed out if someone overwrites an existing installation by re-installing they will be present on that individual's memory stick still.  If the files don't exist, the rm command will simply complain that those files are not present and the script will continue anyway.  The end user probably won't ever see the error message anyway.  I can remove the delete line, but I guess it'll have to be stated somewhere to at least delete the old libs folder prior to reinstalling if doing so from the previous release.  My goal was simply to attempt a fail-safe update for everyone (if that makes sense).

Cebion:
> I've certainly removed the files from the package, but like you pointed out if someone overwrites an existing installation by re-installing they will be present on that individual's memory stick still. If the files don't exist, the rm command will simply complain that those files are not present and the script will continue anyway. The end user probably won't ever see the error message anyway. I can remove the delete line, but I guess it'll have to be stated somewhere to at least delete the old libs folder prior to reinstalling if doing so from the previous release. My goal was simply to attempt a fail-safe update for everyone (if that makes sense). Setting it to do a check if the files exist prior to deletion would be fine, but I was also trying desperately to keep the shell script short, too.

```
[ -f "$GAMEDIR/lib/libasound.so.2" ] && rm -f $GAMEDIR/lib/libasound.so.2 $GAMEDIR/lib/libvorbis.so.0 $GAMEDIR/lib/libvorbisenc.so.2 $GAMEDIR/lib/libvorbisfile.so.3
``` 

😃

Slayer366:
I see that it only checks for the one file, but that's good enough for me.  I'll make this change rq.

Slayer366:
@Cebion @kloptops Lemme know if all's well now

Cebion (approved):
:+1:

PR #101, opened 2024-02-15
Update Port: Billy Frontier for non OpenGL devices

Cebion:
Port no longer needs OpenGL devices.


tabreturn (approved):
Nice to have this running on non-OpenGL devices :)

kloptops (approved):
👍

PR #113, opened 2024-02-20
GMU Music Player - clean up files & libs, add missing libs, update config & sh script, ensure working on JelOS

Slayer366:
Remove unneeded files
Cleanup libs folder
Add missing libs (libmodplug.so.1, libopenmpt.so.0)  
Update configuration file (this was necessary so it would launch on JelOS)
Clean up and update shell script/launcher
Ensure functionality with JelOS
(Yes, I'm aware that JelOS has GMU built into the firmware, but I noticed that on my RG351P with JelOS that OGG files would not play.  My build of GMU now does play OGG/OGA tracks even on JelOS!)


Cebion (approved):
Looks good!

kloptops (approved):
👍

Issue #115, opened 2024-02-21
Port Issue: Cave Story Evo


JeodC:
Cave Story Evo is pretty outdated with the launchscript and oga controls. Spelunky needed a gptokeyb update to accommodate new devices a while ago; this port needs the same treatment. Controls are mapped incorrectly on the Magic XU10.

https://github.com/PortsMaster/PortMaster-New/pull/68

JeodC:
Pull Request #135 recompiled the nxengine-evo binary, which according to some users removed the 720x720 resolution. Need to test the old build against the new build on the RGB30 and verify if the 1:1 aspect ratio was a thing or if the game used a higher resolution that shrunk to fit the screen.

szelek:
I don't know if it is the right place to add input about Cave Story controls, but on R36s handheld there are some issues with controls mapping. L2 and R2 are both mapped to C button, Start is recognized as double clicked, D-Pad Left and Right are mapped as direction buttons (which is good) and as prev. and next weapon button (not good).
I am running latest ArkOS from https://github.com/AeolusUX/ArkOS-R3XS
Port Master is updated to the latest version and Cave Story is downloaded directly from Port Master.

SerpicoSal:
@JeodC Please let me know if you need additional help testing on a RGB30 and I'll be happy to help.

JeodC:
> @JeodC Please let me know if you need additional help testing on a RGB30 and I'll be happy to help.

I don't own one so go for it.

mhoogcarspel:
I don't know how far back we need to go to see this working differently on 720x720 but I tried on latest and commit (3f4a3e86208e742eb2e1cf094ccc56a0a6dd0e19) just before they restructured the renderer in 2.6.5

https://github.com/nxengine/nxengine-evo/releases/tag/v2.6.4

binary from before the last update 
https://github.com/PortsMaster/PortMaster-New/blob/884f91a9419705cf29fa15304c063ea33916bfe6/ports/cave.story-evo/nxengine-evo/nxengine-evo

also just has a 640x480 in the top left corner on RGB30
![2024-04-18-13-48-16-897156914](https://github.com/PortsMaster/PortMaster-New/assets/28956861/f05e7d13-fb52-4a86-abba-bd889ec83146)

The current code doesn't have code for centering or any code to do non-integer scaling as far as I can see

mhoogcarspel:
Someone actually tried this PR on RGB30 with success
[...]

PR #135, opened 2024-02-27
[cave.story-evo] Update binaries, fix controls

octopuserectus:
Compiled the latest version, added GPTOKEYB mapping, rewrote startup script


octopuserectus:
Should fix #115

JeodC (changes requested):
The launhscript is missing `source $controlfolder/device_info.txt` which is where thw $CFW_NAME variable comes from.

JeodC (approved):
Looking good

Cebion (approved):

wwark:
For RGB30, the video mode 720x720 is not present anymore with this commit.

PR #143, opened 2024-03-02
[UPDATE] Cave Story-Evo

JeodC:
Updates the launchscript and renames the settings files to make more sense and actually work.


Cebion (approved):

kloptops (approved):
👍

PR #147, opened 2024-03-03
[Update] Spearmint Mountain - Fix no music

joyrider3774:
I noticed that Spearmint Mountain contained no music while playing and that other games repackaged the apk file to add ogg music, so i tried it out for Spearmint Mountain and it fixed the music. I based myself on the Stars vs Game script to do exactly what that script did (as that one did have music and a repack of the apk).

I'm not sure if the game was supposed to load the music from the gamedata folder but at least on my RG35XX plus it did not do this.


kloptops (approved):
👍

Fraxinus88 (commented):
This works

PR #179, opened 2024-03-11
Add love 11.5 armhf and adapt port to multiarch (armhf)

Cebion:
Add love 11.5 armhf and adapt port to multiarch (armhf)


kloptops (approved):
👍

PR #195, opened 2024-03-19
Fix Shipwreck with newest version from discord

Cebion:
Fix Shipwreck 

https://discord.com/channels/1122861252088172575/1166854578784972812/1202272997688487946


kloptops (approved):
👍

PR #196, opened 2024-03-20
Update Stardew Valley instructions

JeodC:
Stardew Valley compatibility branch was updated to v1.6 today. This uses .NET 6 which is currently broken in the port. Update the port.json instructions to use 1.5.6 until this is fixed (if ever).


kloptops (approved):
👍

PR #203, opened 2024-03-22
Fix angband for 854x480

mxmgorin:
Fix game crash for 854x480 resolution, the font (12x24) was too big for the resolution


kloptops (approved):
👍

PR #213, opened 2024-03-23
Fixes by Kompaan

Fraxinus88:
Fixes to the bash script by Kompaan.
This will exit if it doesn't manage to patch and will display in the log to update portmaster.
I should probably include this in my games with xdelta3...

Also added gameinfo.xml


kloptops (approved):
👍

PR #222, opened 2024-03-25
Update Hexen 2

Slayer366:
Update Hexen II launch scripts to accommodate device combination compatibility through PortMaster detection.
Add libs.aarch64 folder also with fewer libs (credit Cebion).
Add gl4es.aarch64 folder to separate GL4ES for firmwares that do/don't need GL4ES (credit Cebion).
Update port.json - add Cebion as porter.


Cebion (approved):
Looks good now, and tested in discord =)

PR #225, opened 2024-03-25
Update Port: Add armhf support

christopher-roelofs:
# Update port

- add armhf support.
- split libs folders for aarch64 and armhf
- rename binaries because of issues with killing process
- rename gtpk
- update readme to point to repo with portmaster changes for port


kloptops (approved):
👍

PR #240, opened 2024-03-30
Update cannonball-lr port.json

monkeyx-net:
It looks as though if no arch set in port.json then it shows up for x86_64 and I am guessing other arch like armhf?

Test to see if this fixes.

If it does I will change the rest that are missing 

See screenshot here for list of affected ports

https://discord.com/channels/1122861252088172575/1223212073358594058/1223579038996500530


Cebion (approved):
You are correct, cross reference 2048 for an example that I redid for multiarch.

PR #241, opened 2024-03-30
Maninc miner update

monkeyx-net:
Added new sh file and support for armhf and x86_64


Cebion (dismissed):

monkeyx-net:
Closed as needs a new header file as well now

PR #265, opened 2024-04-05
update .sh files for armhf ports with PORT_32BIT="Y"

mhoogcarspel:
update .sh files for armhf ports with PORT_32BIT="Y", some cleanup of port.json to mark gmloader ports as armhf only

PR incoming for PortMaster-GUI to go with as well

### DIFF
--- a/ports/savant/port.json
+++ b/ports/savant/port.json
@@ -22,7 +22,7 @@
     "runtime": null,
     "reqs": [],
     "arch": [
-      "aarch64"
+      "armhf"
     ]
   }
 }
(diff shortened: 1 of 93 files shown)


ports/savant/port.json
kloptops, line 1: Another one without `export PORT_32BIT="y"`

kloptops (changes requested):
Checked them all, you missed 3 ports after you updated the port.json files?

mhoogcarspel:
> Checked them all, you missed 3 ports after you updated the port.json files?
thanks for catching that! 

(added one more sneaky one)

kloptops (approved):
👍

PR #300, opened 2024-04-16
- Update README.md

tabreturn:
See "Extras" section here (which has a Shovel Knight listing) http://portmaster.games/detail.html?name=undertale

I removed the folder structure seeing the Undertale port isn't complicated, anyhow.


Slayer366 (approved):

kloptops:
This was only added this week to avoid confusion.l and so that people can see what the layout is expected to be.

tabreturn:
> This was only added this week to avoid confusion.l and so that people can see what the layout is expected to be.

Should I add the actual Undertale directory structure instead?

kloptops:
Actually yeah I put it in the wrong readme. My bad. 🤦‍♂️

kloptops (approved):
👍

PR #312, opened 2024-04-22
Fix labyrinthoflegendaryloot for trimui

mxmgorin:
The game was crashing when starting on TrimUI Smart Pro. Changes:
- removed libs which were pulling missing dependencies on trimui (these libs also don't needed on other cfw, works on rocknix, jelos, arkos)
- removed setting of custom system mouse cursor in the game code - calls for set/create cursor is crashing on trimui (cursor isn't visible on other cfw and don't need for gameplay)


mhoogcarspel (approved):
Looks good, tested ok on ArkOS and JELOS

Cebion (approved):

kloptops (approved):
👍

PR #316, opened 2024-04-23
Fix cannonballst for trimui

mxmgorin:
Trimui doesn't support shorthand  '|&' operator for redirecting stdout and stderr; so fails to start because of that. Changed to using explicitly 2>&1.

### DIFF
--- a/ports/cannonball-st/Cannonball-st.sh
+++ b/ports/cannonball-st/Cannonball-st.sh
@@ -33,7 +33,7 @@ echo "Loading, please wait... " > /dev/tty0
 
 $ESUDO chmod +x "$GAMEDIR/cannonball"
 
-./cannonball |& tee log.txt /dev/tty0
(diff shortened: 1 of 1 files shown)


Cebion (commented):
We're not doing the piped log output anymore in new ports.
This is what we do now 

### Log the execution of the script, the script overwrites itself on each launch``` 
> "$GAMEDIR/log.txt" && exec > >(tee "$GAMEDIR/log.txt") 2>&1```

mxmgorin:
> We're not doing the piped log output anymore in new ports. This is what we do now
> 
> ### Log the execution of the script, the script overwrites itself on each launch
> ```
> > "$GAMEDIR/log.txt" && exec > >(tee "$GAMEDIR/log.txt") 2>&1
> ```

Changed

Cebion (approved):

kloptops (approved):
👍

PR #324, opened 2024-04-25
MineSector: Set D-Pad to move mouse, add description to gameinfo.xml, add armhf

Slayer366:
Set D-Pad to move mouse for analogless devices such as RG35XX PLUS.
Add description to gameinfo.xml.
Add armhf support
Update shell script and port.json to reflect update


kloptops (approved):
👍

PR #336, opened 2024-04-26
Hexen 2: Add armhf and x86_64, update port.json, gamelist.xml, and scripts

Slayer366:
Add support for armhf and x86_64
Fix cfg not loading on muOS
Update gameinfo.xml

To clarify the config not loading, the symlink to the conf folder was being created as it should, but would still fail to either find or read the contents of the $home/config directory.  Hexen 2 doesn't have a command line option to point it directly to a custom config folder like blake stone does, but does use the autoexec.cfg method that all the Quakes have support for.  By having the script copy the existing config.cfg to the game's autoexec.cfg worked like a charm.  Any OSes that don't have issues with the home folder will still allow users to make and keep changes as it will copy them from the config.cfg file from the last time the game was run so the user won't lose those settings.  This only won't be true for muOS atm.


Cebion (dismissed):

Slayer366:
Sorry, @Cebion.  I see you had approved it for me already.  It's all set and ready to go now this time for sure 👍

Cebion (approved):

kloptops (approved):
👍

PR #341, opened 2024-04-28
Fix noparachute genre casing

mxmgorin:
The game still doesn't show on the site, so I rechecked to ensure everything is correct. I noticed that only the genre name has different casing compared to other ports and the site generator. I don't know if it really makes a difference, but I fixed it just to be sure.


ports/noparachute/port.json
```diff
@@ -14,7 +14,7 @@
         "desc": "No Parachute is a fast-paced arcade game with a pixel art style, in which you will dodge various obstacles in free fall. The man jumped out of a plane, and his parachute didn’t open. But instead of crashing, he fell into a mysterious hole in the ground. He’ll have to fly through a lot of mines and tunnels and meet unexpected visitors. As the game progresses, the unsuccessful skydiver will have to use all his skill and reaction to survive. You will have very little time to assess the situation and make the right decisions, so be prepared to engage your entire concentration. Try to catch Zen (without burning your ass)!",
         "inst": "Purchase and download the game https://store.steampowered.com/app/1575300/No_Parachute/ and copy the exe file into the port folder."
```
kloptops, line 15: Missing comma at the end of this line is the cause of the error. Not sure why the port checker didnt pick this up.

kloptops (changes requested):

kloptops (approved):
👍

PR #343, opened 2024-04-29
Ken's Labyrinth: Update libs for SteamDeck - add sdl12-compat

Slayer366:
Update libs and add new missing after steamOS update.
Add sdl12-compat for overlay radial menu - Steam overlay menus do not work with SDL1 as was discovered with KL.
Since using the wrapper is inconsequential and should work in all use cases, seems best to implement it indefinitely.


kloptops (approved):
👍

Cebion (approved):

PR #360, opened 2024-05-05
Update screenshot.png, cover.png as well as gameinfo.xml

Fraxinus88:
Update the port with a better screenshot/cover as well as gameinfor.xml


Cebion (dismissed):

kloptops (dismissed):
👍

PR #390, opened 2024-05-12
Fix incorrect ScreenWidth for 4:3 devices

gPhantasm:
360px is too wide, causes artifacts on the right side. 320px is perfect.


Cebion:
Has this been tested on the discord on all cfws so that no regression happens?

Cebion:
Close because of Inactivitry.

PR #393, opened 2024-05-14
Change CD Path position on gl4es games to allowd PWD mapping in libgl…

Cebion:
…_CFWname to work


Slayer366 (approved):
Looks like a reasonable solution to me 👍

kloptops (approved):
👍

PR #406, opened 2024-05-17
Rewrote instructions to streamline it with simplified method.

Goy288:
Discovered upon experimentation that the "game.unx" file that Undertale Linux uses can simply just be replaced with the "data.win" file that Undertale Windows uses. Seeing as how this pretty much negates the need for a Linux version, I've rewritten the instructions to help streamline the process for non-Linux users.


kloptops (approved):
👍

Cebion (approved):

PR #434, opened 2024-05-25
Fix quit to main menu for maxdownforce on digital control mode

mxmgorin:
The game uses gptk mapping on digital controls mode (this mode is used for devices with less than 2 sticks because native gamepad support requires 2 sticks) and it is impossible quit to menu on it. To fix this I removed gamepad mapping for SELECT buttons from the game (to prevent double command input) and mapped it to esc


kloptops (approved):
👍

PR #436, opened 2024-05-27
Update chargekid.pck

Fraxinus88:
Optimised game to reduce slowdown due to portals.
Still not perfect but reduces slowdown from level 13 onwards


kloptops (approved):
👍

PR #452, opened 2024-06-09
Fix Serious Sam issue on ROCKNIX, add control scheme for trigger devices

Slayer366:
-Fix Serious Sam black screen on ROCKNIX
-Add control scheme for devices with triggers
-Removed LIBGL_ES setting so it defaults to 2 - minor graphical anomalies occur with this setting, but isn't game breaking and seems to work across all configurations.

### DIFF
--- a/ports/serioussam-tfe/Serious Sam - The First Encounter.sh
+++ b/ports/serioussam-tfe/Serious Sam - The First Encounter.sh
@@ -50,7 +50,6 @@ export SDL_GAMECONTROLLERCONFIG="$sdl_controllerconfig"
 
 if [ "$LIBGL_FB" != "" ]; then
   export SDL_VIDEO_GL_DRIVER="$GAMEDIR/gl4es/libGL.so.1"
-  export LIBGL_ES=1
   export LIBGL_GL=14
 fi
 
--- a/ports/serioussam-tfe/Serious Sam - The First Encounter.sh
+++ b/ports/serioussam-tfe/Serious Sam - The First Encounter.sh
@@ -14,22 +14,26 @@ fi
 
 source $controlfolder/control.txt
 source $controlfolder/device_info.txt
-[ -f "${controlfolder}/mod_${CFW_NAME}.txt" ] && source "${controlfolder}/mod_${CFW_NAME}.txt"
 
 export PORT_32BIT="Y"
+[ -f "${controlfolder}/mod_${CFW_NAME}.txt" ] && source "${controlfolder}/mod_${CFW_NAME}.txt"
 
 get_controls
 
-GAMEDIR=/$directory/ports/sstfe/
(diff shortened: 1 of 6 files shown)


ports/serioussam-tfe/Serious Sam - The First Encounter.sh
```diff
@@ -41,18 +45,19 @@ else
...
-export SDL_VIDEO_GL_DRIVER="$GAMEDIR/gl4es/libGL.so.1"
-export LIBGL_ES=1
-export LIBGL_GL=14
-fi 
-
+export LD_LIBRARY_PATH="$GAMEDIR/Bin/libs:/usr/lib32:$LD_LIBRARY_PATH"
 export SDL_GAMECONTROLLERCONFIG="$sdl_controllerconfig"
 
-$GPTOKEYB "ssam-tfe" -c "$GAMEDIR/serioussam.gptk" &
+export SDL_VIDEO_GL_DRIVER="$GAMEDIR/gl4es/libGL.so.1"
```
Cebion, line 51: We have the put the export SDL_VIDEO_GL_DRIVER in the IF statement else Panfrost will export the gl4es as a driver.

ports/serioussam-tfe/Serious Sam - The First Encounter.sh
```diff
@@ -41,18 +45,19 @@ else
...
+#if [ "$LIBGL_FB" != "" ] && [ "$CFW_NAME" != 'ROCKNIX' ]; then
```
Cebion, line 52: You're overwriting the Rocknix behaviour from https://github.com/PortsMaster/PortMaster-GUI/blob/main/PortMaster/libgl_ROCKNIX.txt

Mind that Rocknix also has a Libmali switch which then requires the exports.

Cebion (changes requested):

ports/serioussam-tfe/sstfe/serioussamtriggers.gptk
```diff
@@ -0,0 +1,40 @@
...
+
+r3 = home
+
+//deadzone_triggers = 0
+mouse_scale = 6144
+
+mouse_delay = 16
+
+//deadzone_y = 2100
+//deadzone_x = 1900
```
kloptops, line 40: GPTK comment format is `#` not `//`, also what is different about the x55 and 353p vs any other device that they need a different controller mapping?

Slayer366: Physical triggers.  Swapping the R1 bumper for the R2 trigger for fire, for example since the layout has the triggers positioned in the "natural" place on the controller.

kloptops: Ahhh okay. Awesome. 😁

Cebion (approved):

PR #466, opened 2024-06-14
Replaced all '|&' occurrances with more widely supported '2>&1 |'

sirchia:
Some handhelds that don't have a 'real' bash shell but some other sh compatible shell don't support the shorthand notation of '|&'.

Change to the equivalent but more widely supported '2>&1 |' notation.


JohnnyonFlame:
Sure, but like, _what_ device? Is this OFW/CFW? I would focus on providing proper bash instead if you control the firmware.

konomikitten:
I'm pretty well versed in shell coding, I've looked over all the changed files and they look good to me. While I'd like to test the changes I don't have a device I was making the earlier commit to fix things for someone else. I also don't own most of the games involved.

PR #468, opened 2024-06-15
Add Catacomb (1989)

Slayer366:
- # New Port for Catacomb (1989)

## Game Information
- **Title**: Catacomb (1989)
- **URL**: [Modified source to work with 1989 game](https://github.com/Slayer366/CatacombSDL-hh)   Since this repo does not reflect the changes made to the source that makes this first release work, I have included this modified source code in a compressed archive which is less than 50KB in size!

### CFW Tests
- [x] AmberELEC
- [x] ArkOS
- [x] ROCKNIX
- [x] muOS

### Resolution Tests
- [x] 480x320
- [x] 640x480
- [x] 720x720 (RGB30)
- [ ] Higher resolutions (e.g., 1280x720)

It may not be much of a concern to test higher resolutions since SDL2 is set to dynamically scale the internal resolution from 320x200 and theoretically should work on most any display res, but it would be nice to have someone to test it anyhow.


kloptops (approved):
👍

Cebion (approved):

PR #485, opened 2024-06-24
Fix minetest for AmberELEC and muOS 

tiopex:
- Remove libcurl.so.4 from libs dir
- Fix extraction archive file
- Enable loopback interface (for self-hosting)


Cebion:
-Tested from a fresh install works good now.

PR #502, opened 2024-07-02
Randungeon/Red Trees Port.json update fix

Tlaggypyro:
i updated both json think i did it right someone correct me


Cebion (changes requested):
see Discord messages

PR #529, opened 2024-07-13
Quake2:  OpenGLES-ControlMappings-NewPlayerModels

Habbening:
Primarily optimised for the RG35xx-H and similar hardware for KeyMappings
Performance optimizations via cfg file
Folder setup for cinematics, music, maps and some additional player models
Including OpenGLES3


ports/quake.2/Quake 2.sh
```diff
@@ -1,4 +1,5 @@
 #!/bin/bash
+# PORTMASTER: quake.2.zip, Quake 2.sh
```
Cebion, line 2: - This is autocreated by portmaster

ports/quake.2/Quake 2.sh
```diff
@@ -12,16 +13,21 @@ else
...
+
+# Function to get controls from control.txt
 get_controls
 
 GAMEDIR="/$directory/ports/quake2"
 
-if [[ $LOWRES == 'Y' ]]; then
-  swidth="320"
+
+ if [[ $LOWRES == 'Y' && $CFW_NAME != "muOS" ]]; then
```
Cebion, line 27: With this you're overriding all 3:2 devices config.
Why hardcode a resolution to a cfw?

Just do `swidth=$DISPLAY_WIDTH` and `sheight=$DISPLAY_HEIGHT` 
to get resolutions from each device.

ports/quake.2/Quake 2.sh
```diff
@@ -36,21 +42,18 @@ else
...
-# fi
+export SDL_GAMECONTROLLERCONFIG="$sdl_controllerconfig"
 
 cd $GAMEDIR
 $ESUDO chmod 666 /dev/tty1
 $ESUDO chmod 666 /dev/uinput
-$GPTOKEYB "quake2" -c "$GAMEDIR/quake2.gptk" &
-SDL_GAMECONTROLLERCONFIG="$sdl_controllerconfig" ./quake2 +set vid_renderer soft +set r_customwidth $swidth +set r_customheight $sheight +set r_hudscale $sscale +set r_menuscale $sscale -datadir $GAMEDIR  +set vid_fullscreen 2 -portable 2>&1 | tee $GAMEDIR/log.txt
+# $GPTOKEYB "quake2" -c "$GAMEDIR/quake2.gptk" &
+$GPTOKEYB "quake2" &
```
Cebion, line 51: gptk was assigned to delete and escape on start and back.
Is this no longer needed?

ports/quake.2/Quake 2.sh
```diff
@@ -36,21 +42,18 @@ else
...
+SDL_GAMECONTROLLERCONFIG="$sdl_controllerconfig" ./quake2 +set vid_renderer gles3 +set r_customwidth $swidth +set r_customheight $sheight +set r_hudscale $sscale +set r_menuscale $sscale -datadir $GAMEDIR  +set vid_fullscreen 2 -portable 2>&1 | tee $GAMEDIR/log.txt
```
Cebion, line 52: Adding gles3 as an option might have impact on all other cfws. Has to be tested.

ports/quake.2/Quake 2.sh
```diff
@@ -36,21 +42,18 @@ else
...
+
 $ESUDO kill -9 $(pidof gptokeyb)
+unset SDL_GAMECONTROLLERCONFIG
```
Cebion, line 55: We don't unset that setting. That was done by a few individuals that is no longer in use.
Example for an up to date start script.

https://portmaster.games/packaging.html#basic-launchscript

ports/quake.2/quake2/gamecontrollerdb.txt
Cebion, line 1: Any reason for a added gamecontrollerdb.txt to the game?
Unless absoluetly necessary we try to avoid a gameocontrollerdb.
PortMaster has its own copy and added each device there is in that file with the correct mapping. This way we ensure all players have the correct mappings no matter what cfw or device.

This is very hardcoded and if a new device get's added this file will be obsolete and will need manual replacement.

ports/quake.2/quake2/quake2.gptk
```diff
@@ -1,24 +1,28 @@
...
+b = space
+x = quote
+y = enter
+l1 = mouse_right
+l2 = =
+l3 = g
+r1 = mouse_left
+r2 = /
+r3 = h
+guide = esc
```
Cebion, line 13: Most devices don't have a guide button.
So relying on putting guide as menu button is bad practice and will fail on most devices.
Either implement a hotkey approach for back + hotkey or don't use the guide button at all.

Many devices also use the "guide" button as R3 for example when they don't have real analog sticks.

ports/quake.2/quake2/ref_gles3.so
Cebion, line 1: adding new libraries or binary requires the game to be retested on all CFWs, I'm afraid.
We can't simply replace these.

Cebion (changes requested):
I'm sorry for these many requests changes but need to be sure that everything works and is fine. 

Quake 2 has over 7.5k downloads.

ports/quake.2/quake2/ref_gles3.so
Slayer366: This is necessary for the new GLES3 API that was added to the Yamagi Quake 2 source port.  This isn't a library, but one of the several supported APIs by the source port that have always been external from the executable.  
ref_gl1.so = OpenGL 1.x
ref_gl3.so = OpenGL 3.x
ref_soft.so = Software mode
ref_gles3.so = GLES 3
Yamagi Quake 2 is build this way so that the renderers can be updated and compiled separately if need be and not required to be integrated into the Yamagi quake2 executable.  If you were to delete all 4 of these, Yamagi quake 2 would be the only thing affected and would never start.  It's a package deal.  They're specific and proprietary to Yamagi Quake 2 and all the supported APIs by the YQ2 executable need to be in the same directory with the binary.  To clarify; these 4 .so files have nothing to do with the underlying Linux OS other than they were compiled for Linux in the same exact way the source port was and they are soley part of the source port.

ports/quake.2/quake2/gamecontrollerdb.txt
Habbening: I've deleted the file.

Habbening (commented):
done per the requests

ports/quake.2/Quake 2.sh
Habbening: removed

ports/quake.2/Quake 2.sh
Habbening: removed

ports/quake.2/Quake 2.sh
Habbening: removed

ports/quake.2/Quake 2.sh
Habbening: .sh updated

Habbening:
Going to close this because slayer366 will resubmit

PR #542, opened 2024-07-17
F-Zero Pocket

MadShmupper:
# New Port for { F-Zero Pocket }

## Game Information
- **Title**: { F-Zero Pocket }
- **URL**: { https://linky439.itch.io/f-zero-pocket }

### CFW Tests
Ensure your game has been tested on all major CFWs:
- [x] AmberELEC
- [x] ArkOS
- [x] ROCKNIX
- [x] muOS

### Resolution Tests
Test all major resolutions:
- [x] 480x320
- [x] 640x480
- [x] 720x720 (RGB30)
- [x] Higher resolutions (e.g., 1280x720)

## File Structure
- Your port should have the following structure:
  - portname/
    - port.json
    - README.md
    - screenshot.jpg
    - cover.jpg
    - gameinfo.xml
    - Port Name.sh
    - portname/
      - <portfiles here>

## Additional Resources
For an in-depth guide on creating a pull request, refer to: [PortMaster Game Packaging Guide](https://portmaster.games/packaging.html#creating-a-pull-request)

PR #545, opened 2024-07-17
Make f-zero pocket rtr

JeodC:
Piggybacking on #542, make f-zero pocket true RTR by skipping the installation step.


kloptops (approved):
👍

PR #547, opened 2024-07-17
Update Descent Trilogy

JeodC:
- Descent 1 & 2 now have control scheme to match Descent 3
- All three ports now require at least one analog stick
- Readmes updated to reflect new controls


kloptops (approved):
👍

PR #559, opened 2024-07-26
Fixes for Descent and Mystik Belle

JeodC:
- Descent 1 had mouse mapping in gptokeyb, other Descent ports did not
- Mystik Belle would not render properly with the libGL.so.1 that was present, use the previous version


ports/mystikbelle/mystikbelle/gl4es/libGL.so.1
Cebion, line 1: I only see the one libgl removed but not replaced

Cebion (changes requested):

ports/mystikbelle/mystikbelle/gl4es/libGL.so.1
JeodC: It's replaced. Binary files aren't shown.

Cebion (approved):

kloptops (approved):
👍

PR #560, opened 2024-07-26
Refactor Undertale

JeodC:
Refactors Undertale to use gmloader instead of box86. Performance should be slightly better, things are less prone to break, and we can use both Windows and Linux versions (tested).

[x] ArkOS
[x] AmberELEC
[x] Rocknix
[x] muOS
[x] Other

* Multi-language support was requested, but this requires Undertale v1.001. This older version uses a different gmloader wrapper, so to reduce complexity, and due to lack of documentation at the time, it was decided multilanguage should be a separate port.


ports/undertale/port.json
```diff
@@ -9,19 +9,16 @@
...
         "porter": [
-            "krishenriksen"
+            "Jeod"
         ],
-        "image": {},
+        "image": null,
         "rtr": false,
         "runtime": null,
-        "reqs": [],
-        "arch": [
```
Cebion, line 23: armhf arch still applies since 32 bit only.

Cebion (dismissed):

Cebion (approved):

ports/undertale/port.json
```diff
@@ -19,6 +19,7 @@
         "image": null,
         "rtr": false,
         "runtime": null,
-        "reqs": ["armhf"]
+        "reqs": [],
+		"arch": ["armhf""]
```
kloptops, line 24: Syntax error, extra `"` here.

kloptops (changes requested):

kloptops (approved):

PR #565, opened 2024-07-29
SonicTT Update #1

JeodC:
- Exclude arkos
  - If used anyway will use extra libs2 folder
- Fix no audio on amberelec due to using extra lib that only arkos needs
- Spell JanTrueno's name right


ports/sonictt16b/Sonic Triple Trouble - 16-Bit.sh
```diff
@@ -25,10 +25,14 @@ cd $GAMEDIR
...
-export LD_LIBRARY_PATH="$GAMEDIR/libs:$LD_LIBRARY_PATH"
+if [ $CFW_NAME == "*ArkOS*" ]; then
+    export LD_LIBRARY_PATH="$GAMEDIR/libs:$GAMEDIR/libs2:$LD_LIBRARY_PATH"
+else
+    export LD_LIBRARY_PATH="$GAMEDIR/libs:$LD_LIBRARY_PATH"
+fi
 
 echo "Loading, please wait... (might take a while!)" > /dev/tty0
-$GPTOKEYB "gmloadernext" xbox360 &
+$GPTOKEYB "gmloadernext" -xbox360 &
```
Cebion, line 35: The command is gptokeyb xbox360 not -xbox360 or so both work?

ports/sonictt16b/port.json
```diff
@@ -24,7 +24,7 @@
     "rtr": false,
     "exp": false,
     "runtime": null,
-    "reqs": [],
+    "reqs": ["!arkos"],
```
Cebion, line 27: You have a if arkos library condition to use other libs. And then you exclude the os completely?

JeodC: Yes because people can be very stubborn.

Cebion (approved):

kloptops (approved):

PR #570, opened 2024-08-01
Remove muOS 32bit checks. Fixes modern muOS versions.

antiKk:
Fixes 2048 port in newer muOS versions since the removal of Retroarch 32bit.

Only muOS specific code has been altered.


kloptops (approved):
LETS GO!

PR #589, opened 2024-08-11
Fix sway fullscreen on some ports

sydarn:
# Fix full screen on sway on some ports

SorR, sonic2013 and OpenJazz.

Tested SorR, but I know the other 2 have the same issue and likely more ports, that will have to be addressed later...

### CFW Tests
- [x ] ROCKNIX

### DIFF
--- a/ports/openjazz/OpenJazz.sh
+++ b/ports/openjazz/OpenJazz.sh
@@ -24,6 +24,10 @@ $ESUDO chmod 666 /dev/tty1
 $ESUDO chmod 666 /dev/uinput
 $GPTOKEYB "OpenJazz" -c "$GAMEDIR/openjazz.gptk" &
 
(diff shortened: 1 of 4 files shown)


sydarn:
There may very well be other ways the apps natively can request fullscreen, but don't. So there may be other ways to fix this issue...

sydarn:
I am drafting this PR, we are trying to examiate if there are more or better options to do this. If you have an idea of how the ports themselves can request through SDL or something, I can test that too.

Leaving the PR open for debate.

sydarn:
Undrafted, we hopefully get a cleaner solution down the road. But for the time being, to get the ports running, we'd like to merge this from ROCKNIX behalf.

kloptops (approved):
Looks good. 👍

PR #596, opened 2024-08-16
Remove usage of arrays from VCMI.sh

ahobsonsayers:
Using arrays causes issues when running VCMI on the Trimui Smart Pro. I am not sure why this is (please tell if you know - im curious!), but I was able to identify the issue from the logs. This PR simply removes usage of bash arrays, and replaces them with alternative but equivalent logic

### DIFF
--- a/ports/vcmi/VCMI.sh
+++ b/ports/vcmi/VCMI.sh
@@ -20,7 +20,7 @@ CUR_TTY=/dev/tty0
 
 PORTDIR="/$directory/ports"
 GAMEDIR="$PORTDIR/vcmi"
-> "$GAMEDIR/log.txt" && exec > >(tee "$GAMEDIR/log.txt") 2>&1
+>"$GAMEDIR/log.txt" && exec > >(tee "$GAMEDIR/log.txt") 2>&1
 
 cd $GAMEDIR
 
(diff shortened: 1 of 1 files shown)


MichalZaniewicz (approved):

ports/vcmi/VCMI.sh
kloptops, line 1: You'll need to do something like this:
```sh
    # Install from gog installer
    EXE_FILE=setup_heroes_of_might_and_magic_3_*.exe
    BIN_FILE=setup_heroes_of_might_and_magic_3_*.bin
    FILES_TO_REMOVE="$EXE_FILE $BIN_FILE"
    BUILDER_OPTIONS="--gog $EXE_FILE"
```

kloptops (changes requested):

ports/vcmi/VCMI.sh
ahobsonsayers: You are right that wildcards don't work in double quotes in bash, but this doesn't seem to be an issue here as the string is being passed to the the `vcmibuilder` which seems to handle the wildcards correctly. I tested this with my GOG version of the game (which includes a build number after the 3) and this worked fine, implying thosnis not a problem.

Regardless, I'm happy to change it if it is preferred! Thanks for looking

Cebion:
Closed due to inactivity

ahobsonsayers:
@Cebion can I get this re-opened? Apologies for being inactive leading to this being closed - my life got busy, and I was happy I just use my custom script.

I am going to make the suggested changed now, and would love to get this merged

ahobsonsayers:
Actually, don't worry, I have opened a new PR here #1091

I will endeavour to get this one merged!

PR #616, opened 2024-08-28
Sonic Ports Update

JeodC:
- Use better handling in the sh files
- Remove Absolute mod (shouldn't have been included)
- Fix Sonic 1 sh (missing variable)


kloptops (approved):
Looks okay. 🐢 🤔

PR #617, opened 2024-08-28
fix camera jumping in minetest for rocknix (bug fix)

DamiansStupidCodeLMAO:
not a new port, just a bug fix for an existing port
honestly i dont know if the checks for rocknix are necessary i just thought they'd be good to include
thanks TheWalkingTaco for finding the problem
i did double check and this does solve the problem, the camera no longer randomly jumps after holding it still for too long


kloptops (approved):
👍

Cebion (approved):

PR #622, opened 2024-08-31
Sonic Ports Update

JeodC:
**Sonic 1 & 2**
- Revert to previous port version (hw render mode on newer rsdk was not functioning on amberelec)
- Leave sonicforever and sonicabsolute binaries available
- Remove Scripts (no longer needed)

**Sonic Mania**
- Settings.ini now sets fsWidth and fsHeight to match screen size
- Reflected in .sh
- Settings.ini now enables controller for debug menu for accessing mods and other things


kloptops (approved):
🤞

Cebion (approved):

PR #628, opened 2024-09-02
Update sonic mania binaries for performance

JeodC:
Built without debug symbols for performance gain on special stages. Updated build guide to reflect.


Cebion (approved):

kloptops (approved):
Nice, only 3 files for a change.

PR #630, opened 2024-09-02
Refactor Spelunky Classic HD

JeodC:
Use gmloader not box86


Cebion (approved):

ports/spelunky/spelunky/port.json
kloptops, line 1: Why is this file added when it should already be in the repo? 🐢

kloptops (approved):
Looks good, but some files are weirdly being added when they already exist on the repo?

PR #654, opened 2024-09-13
Vcmi 1.6.0dev update

ddrsoul:
update VCMI port to latest dev
I`ve uploaded older libs to match my build environment
- gog installer works
- gamepad support works


ports/vcmi/VCMI.sh
```diff
@@ -33,6 +33,11 @@ printf "\033c" > $CUR_TTY
 printf "\033c" > $CUR_TTY
 ## RUN SCRIPT HERE
 
+if [ -f "$GAMEDIR/libs/libicudata.so.63.bz2" ]; then
+    bzip2 -fd "$GAMEDIR/libs/libicudata.so.63.bz2"
+    rm "$GAMEDIR/libs/libicudata.so.63.bz2"
```
kloptops, line 38: This isn't necessary, `bzip2` will extract the file and delete the old one. :)

kloptops (dismissed):

ports/vcmi/VCMI.sh
ddrsoul: you are right! that was unexpected for me even after checking -fd flags
fixed

kloptops (approved):

Cebion (approved):

PR #702, opened 2024-09-29
[Port: Legend of Sword & Fairy]: Add missing reference to the language translation for sdlpal.

feserr:
Add the reference to the language translation for sdlpal because without it the port won't boot as discussed in the Discord thread.


ports/sdlpal/port.json
```diff
@@ -13,7 +13,7 @@
...
-    "inst": "Purchase the game from Steam: https://store.steampowered.com/app/1546570/Sword_and_Fairy/ and put all files into the sdlpal/gamedata folder",
+    "inst": "Purchase the game from Steam: https://store.steampowered.com/app/1546570/Sword_and_Fairy/ and put all files into the sdlpal/gamedata folder. Add the english translation into the sdlpal/gamedata from: https://www.romhacking.net/translations/2441/. If you want to play in other language, modify the variable 'MessageFileName' at the file sdlpal/sdlpal.cfg with the one you want or empty for original language.",
```
JeodC, line 16: Would it be better for the user experience if the launchscript could detect a translation file and add a sed to the configfile?

Cebion (approved):

JeodC (approved):

PR #714, opened 2024-10-01
Fixed tools/build_release.py to allow ports to be moved easily between repos.

kloptops:


Cebion (dismissed):
Do it

kloptops:
Not yet, gotta check to see if i fixed it.

Cebion (approved):
Looks good

Issue #592, opened 2024-08-13
Port Issue: Quake


JeodC:
The [Quake.sh](https://github.com/PortsMaster/PortMaster-New/blob/main/ports/quake/Quake.sh) file needs some attention to account for cfw changes.

- ArkOS has a few forks now, so should use the `*ArkOS*` wildcard check
- MuOS will be allowing retroarch.cfg to live on sd2 on supported devices, which makes the path different. Need to add a nested conditional to check if file exists before using it.

JeodC:
Example with current CFWs used in script. May need to add more such as Knulli and CrossMix.
```bash
#!/bin/sh

XDG_DATA_HOME=${XDG_DATA_HOME:-$HOME/.local/share}

# Determine the control folder
if [ -d "/opt/system/Tools/PortMaster/" ]; then
  controlfolder="/opt/system/Tools/PortMaster"
elif [ -d "/opt/tools/PortMaster/" ]; then
  controlfolder="/opt/tools/PortMaster"
elif [ -d "$XDG_DATA_HOME/PortMaster/" ]; then
  controlfolder="$XDG_DATA_HOME/PortMaster"
else
  controlfolder="/roms/ports/PortMaster"
fi

source $controlfolder/control.txt
source $controlfolder/device_info.txt

get_controls

# Determine RetroArch location and config based on CFW name
case "$CFW_NAME" in
  TheRA)
    raloc="/opt/retroarch/bin"
    raconf=""
    ;;
  RetroOZ)
    raloc="/opt/retroarch/bin"
    raconf="--config /home/odroid/.config/retroarch/retroarch.cfg"
    ;;
  *ArkOS*)
    raloc="/usr/local/bin"
    raconf=""
    ;;
  muOS)
    raloc="/usr/bin"
    # Check if retroarch.cfg exists on sd2
    if [ -f /mnt/sdcard/MUOS/retroarch/retroarch.cfg ]; then
      raconf="-v -f -c /mnt/sdcard/MUOS/retroarch/retroarch.cfg"
    elif [ -f /mnt/mmc/MUOS/retroarch/retroarch.cfg ]; then
      raconf="-v -f -c /mnt/mmc/MUOS/retroarch/retroarch.cfg"
    fi
    ;;
  *)
    raloc="/usr/bin"
    raconf=""
    ;;
esac

GAMEDIR="/$directory/ports/quake"

$GPTOKEYB "retroarch" &
[...]

PR #840, opened 2024-10-23
Update RA Config Path for muOS Banana

duncanyoyo1:
The RA config path changed in muOS Banana, this breaks controls for ports that use libretro cores.

This should fix that.

This should also work with the muOS change to allow RA configs to live elsewhere, as it uses the bind mount. But that has been pushed off to post-Banana anyway so will likely be a while before we need to worry about it.

### DIFF
--- a/ports/2048/2048.sh
+++ b/ports/2048/2048.sh
@@ -28,11 +28,12 @@ elif [[ $CFW_NAME == "ArkOS" ]] || [[ $CFW_NAME == "ArkOS wuMMLe" ]]; then
   raconf=""
 elif [[ $CFW_NAME == "muOS" ]]; then
   raloc="/usr/bin"
-  raconf="--config /run/muos/storage/info/config/retroarch.cfg"
-else
-  raloc="/usr/bin"
-  raconf=""
-fi
+  if [ -f /run/muos/storage/info/config/retroarch.cfg ]; then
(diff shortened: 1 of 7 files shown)


Cebion:
Almost good, but this breaks all non banana versions of the game.
I think maybe a filecheck on the config file could be made.
if file present in location a this config file elese this one.

duncanyoyo1:
Should be good to go now. Hope you don't mind all the commits, I don't have enough disk space to clone the repo so I had to do it all through the GitHub Web UI

Cebion (approved):

JeodC:
This will partially resolve #592 but can we do any more to accommodate other firmwares as well, or do we even need to?

duncanyoyo1:
> Just wondering if replacing where it had
> 
> ```
> -v -f -c 
> ```
> 
> for the arguments in
> 
> ```
> raconf="-v -f -c /mnt/mmc/MUOS/retroarch/retroarch.cfg"
> ```
> 
> with
> 
> ```
> --config
> ```
> 
> is going to be an issue. (I'm guessing it won't be, but figured I'd check)

I actually tested them all and I didn't have any issues. I did have to delete my 2048 save file / state to get it to launch though. No idea why.

Slayer366 (approved):

PR #865, opened 2024-10-27
Consolidate dynamic libraries

JeodC:
- Purges duplicate dynamic libraries with matching md5sums
- Modifies all relevant sh files to point to new PortMaster libstore

Depends on https://github.com/PortsMaster/PortMaster-GUI/pull/178


JeodC:
Tested with f-zero pocket, works. Will test more ports throughout the day.

Cebion:
Note: libgl logic will need to be rewritten in the libgl_cfw if we move them out of gl4es folders.

Cebion:
Best to mark this as draft until it's ready

JeodC:
> Note: libgl logic will need to be rewritten in the libgl_cfw if we move them out of gl4es folders. 

Correct, and all of those are gl4es I believe. They have been consolidated to one version in PortMaster/libs/aarch64.

Cebion:
> > Note: libgl logic will need to be rewritten in the libgl_cfw if we move them out of gl4es folders.
> 
> Correct, and all of those are gl4es I believe. They have been consolidated to one version in PortMaster/libs/aarch64.
The reason we put them into a gl4es folder outside of libs is to only load them on CFW that need them. Rocknix on Panfrost for example has full opengl and if you load gl4es it'll break the port. So in the gui in libgl_rocknix there is a statement to load the the gl4es folder if no panfrost is found else load the lib. So can't live in the same folder.

Slayer366:
> > Note: libgl logic will need to be rewritten in the libgl_cfw if we move them out of gl4es folders.
> 
> Correct, and all of those are gl4es I believe. They have been consolidated to one version in PortMaster/libs/aarch64.

I'm all for consolidating the libs.  This is indeed an excellent idea.  However, this should not be done for GL4ES as with first-hand experience, it's the same reason we don't do it for box86 and box64.  Each game may have its own slightly different compiled version of GL4ES for each game.  I did this very thing for Quakespasm as the one that was tossed to me made 4 frames less on an already low-framerate that it made 6 fps where my compiled version for Quakespasm at the time got 10 fps.  Placing one version of GL4ES for all titles just doesn't seem like the best idea.

JeodC:
Right but I only moved md5sum matching. If we need to we can do sub folders for gl4es versions maybe.

PR #895, opened 2024-11-03
sonictt16b: Fix broken startup on ext4 filesystems.

mkasick:
Commit e811c41697678e37e2512067eb6d1add964bd9f8 breaks sonictt16b when installed on (case-sensitive) ext4 filesystems.

The problem is that the game tries to load language JSON files using titlecase (data/English.json, data/Spanish.json, etc.)¸ but those files are stored in the original .apk as lowercase (data/english.json, data/spanish.json, etc.).  This works fine when the files are stored in game.apk (as they were prior to e811c41697678e37e2512067eb6d1add964bd9f8) since the filename comparisons are case insensitive, but when they're stored bare in the gamedata directory, the loads fail and the game won't boot.

The easiest fix here is to put them back into game.apk, which the updated patchscript does.


JeodC:
Hello, thanks for reporting this. I have created my own simpler solution and will be submitting a pull for that, standby.

PR #972, opened 2024-11-27
Fix Demon of Sakura Pass music

tabreturn:
Add music to .apk


Cebion (commented):
You know that echo only works on soem cfws right? If you want to print something to terminal you can use pm_message =)

You can for example do 

pm_message "Extraction successful."

Cebion (dismissed):

cdeletre (approved):

Cebion (approved):

PR #999, opened 2024-12-02
Refactor UFO 50

JeodC:
Tested and confirmed booting.


Cebion (approved):

cdeletre (approved):

Darky2k1:
This version doesn't work with Audio - It does boot.... Just no Audio.

Darky2k1:
I erased the unzip_audiogroups and compress_audio sections and audio works (Testing on Raspberry Pi with Batocera) - Like the previous version there is still a slight delay in audio Just bad enough to be annoying) - Tried on both directly connected to a PC Monitor / TV (Same Delay)

PR #1001, opened 2024-12-03
Refactor Mystik Belle

### LINKED ISSUE #735: Port Issue: Mystik Belle Box86 -> GMLoader
Mystik Belle is still using box86, but an individual test showed using gmloader offers better performance. Convert it.

JeodC:
Mystik Belle now uses gmloader.armhf instead of box86, drastically increasing performance. Adds a gameinfo.xml and cover.png. Tested install process with Rocknix and muOS.

Resolves #735


Cebion (approved):

Slayer366 (commented):
Should the echo entries be switched out with pm_message in the shell script?

PR #1037, opened 2024-12-12
armagetronad: add missing dependencies

t0b10-r3tr0:


JeodC:
Depending how many ports need it, shouldn't this be something knulli includes, not us?

t0b10-r3tr0:
I have verified that this is working on ROCKNIX, ArkOS, and muOS. @Cebion verified on AmberELEC.

t0b10-r3tr0:
@JeodC, see my comments above. I think it should be brought up if it becomes an issue.

Cebion (approved):

Cebion (approved):

PR #1039, opened 2024-12-12
lierolibre: add missing dependency libFLAC.so.8

t0b10-r3tr0:


t0b10-r3tr0:
I Have verified on ArkOS, ROCKNIX, and muOS. @Cebion verified on AmberELEC.

Cebion (approved):

JanTrueno (approved):
Thanks for the update @t0b10-r3tr0

PR #1052, opened 2024-12-15
Descent 1,2,3: Replace symbolic links with bind mounts.

t0b10-r3tr0:
To enable exFAT support on certain CFWs, We replaced symbolic links with bind mounts using the `bind_directories` function included with PortMaster GUI.

The scope of this task was limited to implementing bind_directories and any fixes preventing the port from launching. Each port was tested in Knulli and one other PM supported CFW (ROCKNIX, most times).

The following criteria were used to validate that a port had passed our testing:
* The port loads without issue, even after rebooting the device.
* The saves/settings persist, even after rebooting the device.
* If port was installed previously, the existing saves/settings were preserved when testing the new version.


ports/alephone-marathon/Marathon.sh
kloptops, line 1: Aren't these in [PR 1050](https://github.com/PortsMaster/PortMaster-New/pull/1050) ?

kloptops (changes requested):

ports/alephone-marathon/Marathon.sh
t0b10-r3tr0: No. Descent and Marathon. They're different developers, platforms, etc. Descent was on DOS, by Interplay(?) and Marathon was on Mac, by Bungie. They're both oddly packaged by 3. Hahaha.

t0b10-r3tr0: It's now fixed. I don't know how those two got mixed up but after merging with upstream, it resolved itself and both PRs are only associated the files related to the PR.

ports/descent3/Descent 3.sh
```diff
@@ -57,7 +56,7 @@ else
 fi
 
 if [ "$LIBGL_FB" != "" ]; then
-  export SDL_VIDEO_GL_DRIVER="$GAMEDIR/gl4es.aarch64/libGL.so.1"
```
Cebion, line 60: What does this change have to do with bind mounts?

Cebion (changes requested):

ports/descent3/Descent 3.sh
t0b10-r3tr0: This was one of the issues we had to address because it prevented the port from starting. So, it was done in the context of the bind-directories fixes. I figured not to make 2 PRs for the change.

Cebion: I know but the title of the PR is replace symbolic links and then does other things :P

kloptops (approved):

Cebion (approved):

PR #1088, opened 2024-12-24
Fix sonic mania display settings

ahobsonsayers:
There are several problems with the display settings for Sonic Mania.

Listing them one by one:

- `awk` does not return a boolean value, so what it returns cannot be used in an if statement. The current if will always use the `else` branch at the moment, therefore. I have fixed this by using an inline `awk` if else
- `pixWidth` in the `grep` and `sed` was set to `PixWidth` with a capital `P` when it needs to be lowercase.
- `Settings.ini` has an uppercase `S`, it was in the script as `settings.ini`, with a lowercase s
- When doing a `sed` to set the `fsWidth` and `fsHeight`, the `pixWidth` line was just being replaced repeatedly (except it wasn't of course, as uppercase `PixWidth` was being used)
  - ~The `Settings.ini` was actually missing `fsWidth` and `fsHeight` - I have added these~

White space changes were made by my shell formatter - but I left them as they seemed reasonable. Happy to revert them if wanted


ports/sonic.mania/sonicmania/Settings.ini
```diff
@@ -27,6 +27,8 @@ exclusiveFS=y
 vsync=n
 tripleBuffering=n
 pixWidth=320
+fsWidth=720
+fsWidth=240
 winWidth=720
 winHeight=240
 fsWidth=640
```
JeodC, line 34: fsWidth exists already

ports/sonic.mania/Sonic Mania.sh
```diff
@@ -35,34 +35,28 @@ $ESUDO chmod 666 /dev/tty1
...
-else
-    WIDTH=$MED
-fi
+WIDTH=$(awk "BEGIN {print ($ASPECT > 1.3 ? $HIGH : $MED)}")
 
-if grep -q "^PixWidth=[0-9]\+" "$GAMEDIR/settings.ini"; then
-    sed -i "s/^PixWidth=[0-9]\+/PixWidth=$WIDTH/" "$GAMEDIR/settings.ini"
-    sed -i "s/^PixWidth=[0-9]\+/fsWidth=$DISPLAY_WIDTH/" "$GAMEDIR/settings.ini"
-    sed -i "s/^PixWidth=[0-9]\+/fsHeight=$DISPLAY_HEIGHT/" "$GAMEDIR/settings.ini"
+if grep -q "^pixWidth=[0-9]\+" "$GAMEDIR/Settings.ini"; then
```
JeodC, line 47: I'm all but certain this comes down to a file system case sensitivity thing. During my refactor tests this worked, so this PR change should be equivalently tested across FAT32, exFAT, and ext4.

ports/sonic.mania/sonicmania/Settings.ini
ahobsonsayers: Oh thanks! Not sure how I missed this below 🤦‍♀️

ports/sonic.mania/Sonic Mania.sh
ahobsonsayers: Yeah I also was pretty confident the Settings.ini not working for me was down to file system case sensitivity on my side, but I guess it still makes sense to use the actual file casing.

As for the grep - this seems like a legitimate problem? File system sensitivity wont affect this right?

JeodC: Yes I concur with grep addition.

ports/sonic.mania/Sonic Mania.sh
```diff
@@ -35,34 +35,28 @@ $ESUDO chmod 666 /dev/tty1
...
 echo "Loading, please wait!" > $CUR_TTY
-$GPTOKEYB "sonicmania" & 
+$GPTOKEYB "sonicmania" &
 ./sonicmania
 
 $ESUDO kill -9 $(pidof gptokeyb)
 $ESUDO systemctl restart oga_events &
-printf "\033c" > /dev/tty1
\ No newline at end of file
+printf "\033c" > /dev/tty1
```
JeodC, line 62:
```suggestion
pm_finish
```

Lines 60+ can be replaced with this single line. :)

JeodC (changes requested):

JeodC:
I tested this on my rp5 and it works well. Good work! Only a few minor tweaks suggested now.

JeodC:
Do you mind terribly if you throw Sonic 1, 2, and CD into this PR as well? They all have a similar display settings modification in sh.

ahobsonsayers:
Absolutely! I'll add that in the next day or so!

ahobsonsayers:
@JeodC -  interestingly, I have looked at Sonic 1, 2, and CD, and I'm not sure if they need any modification. They seem to handle display settings differently, which means they should be OK.

In particular:
 - They **do** have a `settings.ini` with a lowercase `s` unlike Sonic Mania's
 - The if else they use to get `WIDTH` from `ASPECT` is different, as they also support a low resolution, and the conditions use a `bc` to get a boolean value which I am pretty sure is interpretable by bash in the statement (unlike Sonic Mania's awk)
 - The `grep` and `sed` are correct on these games

Looking at these other games, I even wonder if instead of my changes to Sonic Mania, I should just update Sonic Mania to look more like Sonic 1, 2, and CD! I think that would fix Sonic Mania and also keep it closely aligned to these other games. I probably should have thought of this earlier.

Regardless, I have made the changes you suggested - let me know if I should change it again to just align more closely with Sonic CD, 1 and 2.

Thanks for the help!

JeodC (approved):

Cebion (approved):

Maesejesus:
I know this has been "resolved" but I would like to ask for some help.

I have installed Sonic Mania on my Trimui Smart Pro, and it plays at 16:9, but the horizontal resolution is not properly scaled, as there are shimmering pixels. From what I've read, the resolution of the game is 424x240, which would translate into 1272x720, so it goes over by 8 pixels.

I tried to edit the settings.ini but it always revert back to what it was when I open the game. I was told to edit the sonicmania.sh file, but I don't know what to do and when i tried I messed things up and had to reinstall everything. 

Is there a way to avoid that small stretching? I did not want to open a new issue for this as it be my lack of knowledge...

Issue #573, opened 2024-08-02
Port Issue: Downwell Keybinds on TSP are wrong 


Cakeisgreater:
### Device Type

* [ ] RG351P/M
* [ ] RG351V
* [ ] RG351MP
* [ ] RG353M/V/P
* [ ] RG503
* [ ] RG552
* [ ] RGB10
* [ ] RGB30
* [ ] RK2023
* [ ] X55
* [X] Other device (Trimui Smart Pro)
* [ ] All of the above

### Firmware

* [ ] ArkOS
* [ ] ArkOS Wummle
* [ ] AmberELEC
* [ ] JelOS
* [ ] RockNix
* [ ] uOS
* [ ] RetroOZ
* [ ] TheRA
* [X] Crossmix OS 1.2.1
#### Description

Inexpicably, left and right movement is bound to up and down on the right analogy stick, the action key is bound to the "y" button, and the menu is bound to the "trimui" button.

I've tried the port on my MuOS RG35xx H and the controls are bound to more sensible keys (D-pad, etc).

#### Reproduction Steps

Install the port on Trimui Smart Pro running Crossmix OS 1.2.1.

Play Downwell.

JeodC:
@Fraxinus88 Did you want to work this or should it be closed as unplanned?

PR #1093, opened 2024-12-24
Fix downwell keybindings

ahobsonsayers:
As mentioned in  #573 the key bindings for Downwell are very odd.

Firstly, the right stick does the moving, when the left stick should do it. Secondly, you have to press up and down on the right stick to move left and right. This PR fixes these issues, updating Downwell to use sensible key bindings.

I have also updated every button to be mapped to space, so a player can choose whichever button they prefer to play.

### DIFF
--- a/ports/downwell/downwell/downwell.gptk
+++ b/ports/downwell/downwell/downwell.gptk
@@ -6,15 +6,15 @@ down = \"
 left = \"
 right = \"
 
-left_analog_up = \"
-left_analog_down = \"
-left_analog_left = \"
-left_analog_right = \"
+left_analog_up = up
+left_analog_down = down
+left_analog_left = left
+left_analog_right = right
 
-right_analog_up = right
-right_analog_down = left
-right_analog_left = up
-right_analog_right = down
+right_analog_up = \"
+right_analog_down = \"
+right_analog_left = \"
+right_analog_right = \"
 
 l1 = \"
 l2 = \"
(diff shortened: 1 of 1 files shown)


Fraxinus88:
So the key bindings for the right are as such so that Tate mode can be used. The Left-stick and D-pad doesn’t need bindings as it uses the SDL controls I.e gamepad controls for the game

ahobsonsayers:
> So the key bindings for the right are as such so that Tate mode can be used. The Left-stick and D-pad doesn’t need bindings as it uses the SDL controls I.e gamepad controls for the game.
> 
> Every button as a space is not ideal as you can’t select the back button

Thanks for the review!

Ok, the binding for the right for tate mode makes sense I guess, but the left sick and d pad don't work at all without this change. Maybe I leave the right stick as it is, and also add binding for the left stick and d pad then?

As for all the buttons being space - I can see this issue, but back wasn't working before this change anyway - in fact I thought maybe there wasn't a back button at all! Maybe we should explicitly map one or two buttons to back, and leave the other bindings as is

Fraxinus88:
Yup and we can't commit these changes to the repo because it will break all the other CFWs with controller support. What I can think off is to detect Crossmix/TSP and substitute the key bindings into the gptk

Cebion:
Discuss this on the discord. Once you have a solution feel free to make a new PR

PR #1099, opened 2024-12-27
Blood: Correct library path

ben-willmore:
Blood.sh has a hard-coded path in LD_LIBRARY path that doesn't point to the intended folder on all platforms. This should be relative to $GAMEDIR instead.

### DIFF
--- a/ports/blood/Blood.sh
+++ b/ports/blood/Blood.sh
@@ -16,8 +16,8 @@ source $controlfolder/control.txt
 
 get_controls
 
-export LD_LIBRARY_PATH="$LD_LIBRARY_PATH:/storage/roms/ports/Blood/lib:/usr/lib"
 GAMEDIR="/$directory/ports/Blood"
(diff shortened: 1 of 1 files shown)


ports/blood/Blood.sh
```diff
@@ -16,8 +16,8 @@ source $controlfolder/control.txt
 
 get_controls
 
-export LD_LIBRARY_PATH="$LD_LIBRARY_PATH:/storage/roms/ports/Blood/lib:/usr/lib"
 GAMEDIR="/$directory/ports/Blood"
+export LD_LIBRARY_PATH="$LD_LIBRARY_PATH:$GAMEDIR/lib:/usr/lib"
```
Cebion, line 20:
```suggestion
export LD_LIBRARY_PATH="$GAMEDIR/lib:/usr/lib:$LD_LIBRARY_PATH"
```

Cebion (changes requested):

Cebion (approved):

JeodC (approved):

PR #1104, opened 2024-12-28
Daikatana -- add libEGL & libFLAC.so.8, update libGL, bind_directories

ben-willmore:
Fixes for Daikatana on knulli:

* add libEGL (similar to Redneck Rampage and other ports)
* add libFLAC.so.8 (knulli and other recent linuxes only provide .12)
* use bind_directories as proposed by t0b10-r3tr0

Also (not strictly required):
* update libGL to cebion's recent build (in line with Redneck Rampage)
* remove `source $controlfolder/device_info.txt` (in line with t0b10-r3tr0's update)


Cebion (approved):
Thank you, tested and works well now.

kloptops (approved):

PR #1105, opened 2024-12-28
Changes libSDL_sound to a version that does not include speex

bmdhacks:
Speex is an oldschool codec for realtime voice over modems and is not used in any games that I know of. libspeex is not included in knulli, so none of these games will work in knulli because of this dependency. This new libSDL_sound fixes this problem.

I compiled this library on the Docker image found here: https://portmaster.games/docker.html
The source code for libSDL_sound was from here:
https://github.com/Ancurio/SDL_sound rev 04798ba55dccd18b094c0f6a2630c2fe7b15aa86

Before building I removed /lib/aarch64-linux-gnu/libspeex* (because the configure script is broken) and then ran:

```
./configure --prefix="/root/SDL_sound_install" --disable-speex 
make
make install
```


bmdhacks:
I only own a Trimui Smart Pro, but I confirmed this fixes To the Moon on it under knulli.  Also it doesn't break on CrossmixOS but it worked before because Crossmix has speex installed.

kloptops:
Wouldn't it be easier to add libspeex to Knulli?

These ports all need to be tested on various CFW for such a change to be accepted.

bmdhacks:
I can pursure getting speex added to knulli, but honestly nothing has used the speex codec in decades so it seems suboptimal to keep linking it.  For what it's worth, all these ports were already broken on my device, so it's not like we were in a stable state before.  If you'd like I can post some detailed readelf comparisons of the symbols of the two libraries and calling games to show that this change doesn't alter the abi.

kloptops:
I have spoken to acmeplus and it will be included in future builds of knulli. 😁

bmdhacks:
Works for me. I'll close this then?

Cebion:
Just need to make sure any new port uses this new lib ^^

Cebion:
Or I'll test them and we push them through ^^

Cebion (approved):
Tested

Cebion:
Tested works fine

PR #1107, opened 2024-12-29
fix to_lower_case() to work correctly on vfat partitions

bmdhacks:
This was preventing these three games from running on CrossmixOS, which uses a FAT partition for ports.
Tested Fallout 1 after this and it works.

### DIFF
--- a/ports/avp/Alien vs Predator.sh
+++ b/ports/avp/Alien vs Predator.sh
@@ -26,10 +26,13 @@ cd $GAMEDIR
 
 to_lower_case() {
     for SRC in $(find "$1" -depth); do
-    DST=$(dirname "${SRC}")/$(basename "${SRC}" | tr '[A-Z]' '[a-z]')
-    if [ "${SRC}" != "${DST}" ]; then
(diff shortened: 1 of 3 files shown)


Cebion:
Will need to test this later on the other cfws just to be sure =)

Cebion (changes requested):
Works good so far. Only thing now is we don't use echo anymore. 
We use the function pm_message "message"

For understanding we do inside pm when we load control.txt 

```

pm_message() {
    # Usage: pm_message "Some message here."
    if [[ "$CFW_NAME" == "muOS" ]]; then
        /opt/muos/extra/muxstart "$1"
    else
        echo "$1" > $CUR_TTY
    fi
}```

cdeletre:
> Works good so far. Only thing now is we don't use echo anymore. We use the function pm_message "message"

I think it's ok here as it will go into the log.txt, I don't think we want a popup on the screen for this kind of information ?

bmdhacks:
Just to clarify, the usage of echo was already there.  My change only adds a halfway step to not `mv "FOO"  "foo"` on a fat filesystem.

Cebion (approved):

PR #1109, opened 2024-12-30
VHR: Update gmloader.json

JeodC:
This makes it not crash. `os_type` is determined at runtime due to the xdelta and part of the patch requires `os_type` to be android.


JanTrueno (approved):
👍

kloptops (approved):

PR #1129, opened 2025-01-03
sonic CD: Update README.md

plaidman:
update sonic cd readme to point to the popular Miracle Edition mod.


JeodC:
This is an unnecessary mod recommendation, the reason Sonic 1 and Sonic 2 refer to their two mods is because they have special binaries for them. In normal cases such as this, we should probably not endorse any specific mod even if it is the most popular one used.

PR #1131, opened 2025-01-03
sonic 2 updates for S2A mod

plaidman:
- update readme to notify users of the correct version of S2A to use
- update modconfig.ini to show the correct folder name for the most recent S2A mod
- update game launch script to look for the new mod name (preserving the old mod name for backwards compatibility)

### DIFF
--- a/ports/sonic.2/sonic2/mods/modconfig.ini
+++ b/ports/sonic.2/sonic2/mods/modconfig.ini
@@ -1,3 +1,4 @@
 
 [mods]
 S2A=false
+Sonic2Advance=false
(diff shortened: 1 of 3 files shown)


ports/sonic.2/sonic2/mods/modconfig.ini
```diff
@@ -1,3 +1,3 @@
 
 [mods]
-Sonic2Absolute=false
```
JeodC, line 3: For people using older mod versions this name still applies.

plaidman: I think it'll probably be fine if both are kept (though all versions of the mod I've seen are named S2A by default). updated.

JeodC (approved):

Cebion (approved):

PR #1132, opened 2025-01-03
Shadow Warrior: Add libs/libFLAC.so.8

ben-willmore:
Library is missing on knulli, which has libFLAC.so.12


ports/shadow.warrior/shadow-warrior/libs/libFLAC.so.8
kloptops, line 1: Where is this file from?

We have used this version in the past:
- ./ports/pacman/pacman/libs.armhf/libFLAC.so.12
- ./ports/hexen2/hexen2/libs.armhf/libFLAC.so.12
- ./ports/minesector/minesector/libs.armhf/libFLAC.so.12

md5sum: `cf826ddbc3f224307934b9bb74e9c2d6`

kloptops (changes requested):

ports/shadow.warrior/shadow-warrior/libs/libFLAC.so.8
ben-willmore: Thanks for looking at this! It's an aarch64 libFLAC.so.8 that needs to be added not .12. I took one which is present in existing ports (md5 21990670ff58d808a86aef9fee860b09), e.g. duke3d / gatoroboto.

kloptops: Whoops. Not sure how I mixed up the arch! Perfect!

kloptops (approved):

Cebion (approved):
Tested on AmberELEC; works fine

PR #1139, opened 2025-01-04
minetest: Add libcurl that pulls in fewer libraries

ben-willmore:
Minetest does not run on knulli because the included version of libcurl pulls in a lot of libraries that are not present in /usr/lib -- libcrypto.so.1.1, libgcrypt.so.20, libgssapi_krb5.so.2, libgssrpc.so.4, libidn2.so.0,  libidn2.so.4, libk5crypto.so.3, libkdb5.so.10, libkrad.so.0, libkrb5.so.3, libkrb5support.so.0, libselinux.so.1 .

This can be fixed on knulli by just deleting `minetest/libs.aarch64/libcurl.so.4` (and using the system one). However, I imagine this will break it on other CFWs. This PR uses the libcurl.so.4 from moonlightnew, which hopefully should work for everyone.


kloptops (approved):

Cebion (approved):

ben-willmore:
Thanks!

PR #1145, opened 2025-01-05
Descent 3: Add libEGL which is needed on knulli

ben-willmore:
This is similar to recent fixes for Jedi Academy, Jedi Outcase, Daikatana and others.


JeodC:
Needs a test on not-knulli to ensure EGL doesn't break other cfw.

ports/descent3/Descent 3.sh
```diff
@@ -56,6 +56,7 @@ fi
 
 if [ "$LIBGL_FB" != "" ]; then
   export SDL_VIDEO_GL_DRIVER="$GAMEDIR/gl4es.$DEVICE_ARCH/libGL.so.1"
+  export SDL_VIDEO_EGL_DRIVER="$GAMEDIR/gl4es.$DEVICE_ARCH/libEGL.so.1"
```
Slayer366, line 59: Use an if statement to check for Knulli instead so it only gets applied for Knulli
```suggestion
  if [[ "${CFW_NAME^^}" == "KNULLI" ]]; then
    export SDL_VIDEO_EGL_DRIVER="$GAMEDIR/gl4es.$DEVICE_ARCH/libEGL.so.1"
  fi
```

Slayer366 (changes requested):

ports/descent3/Descent 3.sh
Slayer366: Likewise, the presence of libEGL.so.1 has been enough to cause problems on other CFW.  It might be a good idea to have the script move or rename libEGL.so.1 to something else when not running Knulli perhaps.

JeodC: Why does Knulli need EGL in the first place, and if other cfw have it in their `usr/lib`, why doesn't Knulli?

Slayer366: Sounds like Knulli needs to address this IMO

Slayer366:
Disregard my suggestion from earlier.  Testing this across CFWs is recommended, but it may seem that simply adding libEGL.so.1 will not affect it on other firmwares due to how PortMaster initializes GL/GL4ES per configuration.  If it checks out, this PR may require little to no change.

Cebion:
Based on the discussion and Johnny's previous guidance, including libEGL.so.1 alongside libGL.so.1 is the correct approach since:

1. It's part of the gl4es build (compiled with -DEGL_WRAPPER=ON) and belongs with the gl4es files
2. The existing conditional `if [[ "$LIBGL_FB" != "" ]]` ensures the EGL exports only apply where needed
3. The gl4es directory is not exported for mainline devices, preventing conflicts
4. Other ports (Stardew Valley, Jedi Academy, Jedi Outcast, Daikatana) use this same pattern successfully
5. This aligns with Johnny's consistent recommendations for GL4ES implementation

No need for Knulli-specific conditions - the current exports in the script are sufficient:
bash
if [[ "$LIBGL_FB" != "" ]]; then
    export SDL_VIDEO_GL_DRIVER="$GAMEDIR/gl4es.$DEVICE_ARCH/libGL.so.1"
    export SDL_VIDEO_EGL_DRIVER="$GAMEDIR/gl4es.$DEVICE_ARCH/libEGL.so.1"
fi

Testing on the dfiferent cfws of course still needs to happen.

JeodC:
This may be something moot if @lgt2x intends to update the port with the upstream v1.6 changes anyway. A test build of it I tried last month uses gles, not gl4es.

Lgt2x:
Note than 1.6 will likely not be ready until at least a couple of months

JeodC:
I'll handle this, I have some other changes to add to Descent 3 anyway.

PR #1149, opened 2025-01-06
Vortexion: fix freeze when clearing game

Kilvoctu:
This is updated Vortexion gamedata that fixes an issue in which the game will freeze and crash right after beating the final boss


Cebion (approved):
Tested on rgcubexx (knulli, amberelec, and arkos)

kloptops (approved):
Works on RetroDECK.

Got a high score of 1000. 😁

PR #1155, opened 2025-01-07
Barbarian: Make binary executable, move gl4es->gl4es.armhf, add libEGL

ben-willmore:
The Barbarian port is another one that needs the libEGL treatment for knulli.

I also found that after unzipping the game files in gamedata/, the binary needed to be made executable, so I have added `$ESUDO chmod a+x $BINARYNAME`.

Barbarian.sh references `$GAMEDIR/gl4es.armhf/libGL.so.1` but the actual folder is `$GAMEDIR/gl4es` so I renamed the folder to match the script.

The armhf libEGL comes from Shovel Knight, which has already had similar treatment.

Thanks again for looking at these!


JeodC:
I think knulli should just include libEGL.

Cebion:
> The Barbarian port is another one that needs the libEGL treatment for knulli.
> 
> I also found that after unzipping the game files in gamedata/, the binary needed to be made executable, so I have added `$ESUDO chmod a+x $BINARYNAME`.
> 
> Barbarian.sh references `$GAMEDIR/gl4es.armhf/libGL.so.1` but the actual folder is `$GAMEDIR/gl4es` so I renamed the folder to match the script.
> 
> The armhf libEGL comes from Shovel Knight, which has already had similar treatment.
> 
> Thanks again for looking at these!

We decided not do do gl4es.armhf folders because the script  because of

```
 if [ -d "$PWD/gl4es.$DEVICE_ARCH" ]; then
    export LD_LIBRARY_PATH="$PWD/gl4es.$DEVICE_ARCH:$LD_LIBRARY_PATH"
  elif [ -d "$PWD/gl4es" ]; then
    export LD_LIBRARY_PATH="$PWD/gl4es:$LD_LIBRARY_PATH"
  fi```
  
  If you rename the folder to .armhf it will break

The armhf is just a leftover.

Cebion (changes requested):
See comment

Cebion (approved):

kloptops (approved):

PR #1173, opened 2025-01-10
Update Siege of Osaka to CRT-less version

tabreturn:
This is an update for an existing Ready-to-Run port, Siege of Osaka (https://portmaster.games/detail.html?name=siegeofosaka). The original port is quite demanding thanks to the CRT filter, but now there's a CRT-less version (https://hwilson.itch.io/siege-of-osaka) that vastly improves the performance of the port (especially on lower-end devices).

This update entailed switching over to gmloadernext, hence there are a bunch of file changes beyond just the game files.


ports/siegeofosaka/Siege of Osaka.sh
```diff
@@ -1,8 +1,7 @@
-#!/bin/bash
+a#!/bin/bash
```
Cebion, line 1:
```suggestion
#!/bin/bash
```

Cebion (changes requested):

ports/siegeofosaka/Siege of Osaka.sh
tabreturn: Fixed. Thx

Cebion:
Header is still wrong

```
#!/bin/bash

XDG_DATA_HOME=${XDG_DATA_HOME:-$HOME/.local/share}

if [ -d "/opt/system/Tools/PortMaster/" ]; then
  controlfolder="/opt/system/Tools/PortMaster"
elif [ -d "/opt/tools/PortMaster/" ]; then
  controlfolder="/opt/tools/PortMaster"
elif [ -d "$XDG_DATA_HOME/PortMaster/" ]; then
  controlfolder="$XDG_DATA_HOME/PortMaster"
else
  controlfolder="/roms/ports/PortMaster"
fi
```

cdeletre (approved):

PR #1196, opened 2025-01-14
Billy Frontier -- add libEGL

ben-willmore:
Add libEGL to the Billy Frontier port, to make it work on Knulli, as applied elsewhere, e.g. https://github.com/PortsMaster/PortMaster-New/pull/1104

Also update to use pm_finish

Tested on knulli and muOS


ports/billyfrontier/Billy Frontier.sh
```diff
@@ -48,6 +47,6 @@ fi
 $GPTOKEYB "BillyFrontier.${DEVICE_ARCH}" -c "./billyfrontier.gptk" &
 ./BillyFrontier.${DEVICE_ARCH}
 
-$ESUDO kill -9 $(pidof gptokeyb)
-$ESUDO systemctl restart oga_events &
+pm_finish
+
 printf "\033c" > /dev/tty0
```
Slayer366, line 52:
```suggestion
```

Slayer366 (changes requested):

ports/billyfrontier/Billy Frontier.sh
ben-willmore: Thanks for the review! It's unclear to me if this should be removed -- it may have some value in resetting the terminal. Is there a strong reason to remove it?

Cebion: > Thanks for the review! It's unclear to me if this should be removed -- it may have some value in resetting the terminal. Is there a strong reason to remove it?

Yes we don't do this in general. 
Most cfws don't even support clearing the terminal this way. All other stuff is handled in pm_finish

kloptops (approved):

PR #1210, opened 2025-01-15
reminiscence instructions for data and audio files

plaidman:
update reminiscence readme to include instructions about data files and audio files.


plaidman:
The amiga audio files can be found https://www.exotica.org.uk/wiki/Flashback here, but I wasn't sure if I should list it on the readme. It seems like abandonware but I err'd on the side of caution and didn't include the link in case that's not kosher. If you guys think it's OK to include it, I can add a new commit.

ports/reminiscence/README.md
Cebion, line 1: Instructions we put into our port.json files

https://portmaster.games/port-json.html

See the Section "Instructions" which are shown on the device and

"Instructions (markdown) which are shown on the website)

Cebion (changes requested):

ports/reminiscence/README.md
plaidman: all set.

Cebion (approved):

PR #1235, opened 2025-01-24
Adjust Mindustry Java runtime name

binarycounter:
This just updates the runtime name to remove the "aarch64" qualifier, as PM doesn't need it.


cdeletre (approved):

Cebion (approved):

cdeletre (changes requested):
In port.json you don't add the extension squashfs for the runtime

PR #1269, opened 2025-02-05
Update Ubermosh collection

Slayer366:
Clean up and update Ubermosh collection
Update gameselector
Reduce png image sizes
Make case sensitivity consistent
Use .port files for new regimen vs old pseudo .apk files
Reduce .port file sizes by removing unused lib architectures such as MIPS
Rename shell scripts in subdirectory from .sh to .run so not to show up in EmulationStation on some CFWs
Update gameselector.conf to reflect the above change
Update all shell scripts to reflect new PortMaster functionality
Update port.json
Game would crash after several seconds due to RAM usage.  Creating a swapfile fixed this issue when EmulationStation would eat up 300MB+ of RAM particularly on systems with less than 2GB of RAM.  To address this, a swapfile will be created/recreated if on ArkOS or Knulli and ONLY if the device has less than 2GB of RAM.


Cebion (approved):
Good job!

ports/ubermoshcollection/Uber/gamedata/omega.run
```diff
@@ -0,0 +1,65 @@
...
+  controlfolder="/opt/tools/PortMaster"
+elif [ -d "$XDG_DATA_HOME/PortMaster/" ]; then
+  controlfolder="$XDG_DATA_HOME/PortMaster"
+else
+  controlfolder="/roms/ports/PortMaster"
+fi
+
+source $controlfolder/control.txt
+export PORT_32BIT="Y"
+[[ "${CFW_NAME^^}" == 'MUOS' ]] && directory=$(realpath "$0" | sed 's|^\./||' | awk -F'/' '{print "/" $2 "/" $3}')
```
kloptops, line 17: This shouldn't be necessary on newer builds of PortMaster on muOS?

kloptops (approved):

Cebion (approved):

PR #1278, opened 2025-02-09
add cover and correct the path folder

pedrobaratafernandes:
i forget to change the keyword portfolder in -> GAMEDIR=/$directory/ports/portfolder/
add cover


Cebion (changes requested):
You are rearranging the whole .sh
- The port folder location i fixed already 30 minutes ago.
- The godot version was also fixed already
-  you're reintroducing comments

Make a sync of your repo first =)

PR #1280, opened 2025-02-09
VCMI 1.6.5 Update

ddrsoul:
This is update to latest stable VCMI (1.6.5). Built in same environment as previous port, so should not be broken (ubuntu 20.04).
packed small files into one zip, updated launchscript a little


tabreturn:
Closing this for now. It's been in draft status for 8 months. But please reopen if/when necessary :)

PR #1288, opened 2025-02-12
fix sigil.doom file locations

plaidman:
this is the only preconfigured .doom file that has /official/ folder


JeodC (approved):

JeodC:
Good catch, this is the last `.doom` file I added, and my own personal structure keeps official iwads in that subfolder.

Cebion (approved):

PR #1292, opened 2025-02-12
UFO50: Hotfix

### LINKED ISSUE #1290: Port Issue: UFO 50 getting stuck when choosing quit after defeat in "BUG HUNTER" / loseing in magic garden
### Device Type

*  Other device (rg35xxsp)

### Firmware

* muOS

#### Description

Game is stuck in a black screen after exiting "bug hunter" via the quit ingame option.

#### Reproduction Steps

1. Die in "bug hunter".
2. select quit.

LOG:
[log.txt](https://github.com/user-attachments/files/18769252/log.txt)

JeodC:
Since UFO 50 now uses dynamic texture loading, we no longer need to use `draw_texture_flush()` in `scrCleanup`, and if we do, we cannot exit to titlescreen properly due to blackscreen caused by cleaned textures.

Resolves #1290


Cebion (dismissed):

JeodC:
Additional push to add `PortMaster` to the version string to indicate it's modified from the original.

cdeletre (approved):

PR #1313, opened 2025-02-17
Signus: Fix force-quit, use cover.png as main image

ben-willmore:
The version that I originally submitted was missing a fix for non-functional force-quit, which is fixed by this PR. I have tested this on muOS, Rocknix, Knulli. I also corrected the image reference in gameinfo.xml.


ben-willmore:
Also tested on ArkOS

Cebion (approved):

kloptops (approved):

PR #1329, opened 2025-02-23
Add digital tamers reborn

JeodC:
# New Port for Digital Tamers Reborn

### CFW Tests
- [x] AmberELEC
- [x] ArkOS
- [x] ROCKNIX
- [x] muOS

### Resolution Tests
Test all major resolutions:
- [x] 480x320
- [x] 640x480
- [x] 720x720 (RGB30)


ports/digitaltamersreborn/DigitalTamers Reborn.sh
```diff
@@ -0,0 +1,80 @@
...
+> "$GAMEDIR/log.txt" && exec > >(tee "$GAMEDIR/log.txt") 2>&1
+$ESUDO chmod +x -R $GAMEDIR/*
+
+# Exports
+export PATCHER_FILE="$GAMEDIR/tools/patchscript"
+export PATCHER_GAME="$(basename "${0%.*}")" # This gets the current script filename without the extension
+export PATCHER_TIME="2 to 5 minutes"
+export SDL_GAMECONTROLLERCONFIG="$sdl_controllerconfig"
+
+# dos2unix in case we need it
```
kloptops, line 37: This should never be needed, and would need to be fixed at the source.

JeodC: It's for manual installs from GitHub, and harmless to have in there. Doesn't affect load time or anything.

Cebion: Permissions is one thing, but CR LF files do not happen by manually downloading a file. The zip files from our website may loose permissions but never their encoding. This only happens if a file is created under windows for example and then copied over. 

Hence why this should never be needed.

kloptops: If we need this in our scripts, things are far too broken.

If your local git repo is having issues, you need to change your git repo settings to not convert line endings.

JeodC: Never mind that it's already on dozens of others. Fine.

kloptops (approved):
I have a list of the files needing fixing, will get to removing it in the coming days. ❤️

Cebion (approved):

PR #1372, opened 2025-03-07
New port: OpenEnroth

ben-willmore:
# New Port for OpenEnroth

NB The executable is large, and I split into parts using `tools/build_data.py`. The `.gitignore` did not work for me -- it ignored the parts as well as the original file. I think everything is now correct but it is worth checking.

## Game Information
- **Title**: OpenEnroth
- **URL**: https://github.com/OpenEnroth/OpenEnroth

### CFW Tests
Ensure your game has been tested on all major CFWs:
Custom Firmwares:
[X] AmberELEC
[!] ArkOS -- not possible, requires newer gcc/glibc
[X] ROCKNIX
[X] MuOS
[X] Knulli**
[] Crossmix**

Drivers: 
[X] Mali
[X] Panfrost
[X] Adreno (ROCKNIX RP5)

Resolutions: 
[!] 480x320**
[X] 640x480
[] 720x720**
[X] Higher resolutions**

## File Structure
- Your port should have the following structure:
  - portname/
    - port.json
    - README.md
    - screenshot.jpg
    - cover.jpg
    - gameinfo.xml
    - Port Name.sh
    - portname/
      - <portfiles here>

## Additional Resources
For an in-depth guide on creating a pull request, refer to: [PortMaster Game Packaging Guide](https://portmaster.games/packaging.html#creating-a-pull-request)

PR #1377, opened 2025-03-08
Downwell innoextract proof of concept

Doronimmo:
Add innoextract to downwell port (support for installer extraction)
DISCLAIMER: this is a proof of concept - any contributions are welcomed.


ben-willmore:
I did something similar in #1372 and there are a couple of current ports that also bundle innoextract. It massively improves the install experience if you don't have windows. There are probably a large number of ports that would benefit from this.

Cebion:
Was this tested?

Doronimmo:
yup

ben-willmore:
This looks promising so I am reminding you to make a testing thread.

Doronimmo:
> This looks promising so I am reminding you to make a testing thread.

Thanks! added this to the downwell thread, not quite sure how to create a testing thread for somthing like this.

JeodC:
How large is innoextract? If multiple ports will end up using it then it should just become a part of out utilities shipped with the gui.

Doronimmo:
> How large is innoextract? If multiple ports will end up using it then it should just become a part of out utilities shipped with the gui.

Yeah that would be optimal, the PR was originally just a proof of concept for an issue about that :D

Doronimmo:
Huh, I forgot to add innoextract!

ben-willmore:
Innoextract is 1.4MB including the necessary boost libraries. Could this be included in PM tools?

Cebion:
> > Innoextract is 1.4MB including the necessary boost libraries. Could this be included in PM tools?
> 
> That's what I'd prefer to do if ports will be using it. @Cebion?

Why not

ben-willmore:
Closing this because the plan is to add this to portmaster tools, not this port

PR #1379, opened 2025-03-08
AvP: Correct metadata

ben-willmore:
The AVP port is the 1999 Game "Aliens versus Predator" but some of the assets, release date, etc were from the 2010 game. Also the "s" was missing in the port name.


Cebion (dismissed):

ports/avp/Aliens versus Predator.sh
kloptops, line 1: Probably not a good idea to rename this. 😁

kloptops (changes requested):

Cebion (approved):

PR #1383, opened 2025-03-10
Tantibus fix json

NotYourAveragePaladin:
Fixing .json to read "Ready to run"


ports/tantibus/tantibus/libs.aarch64/libspeex.so.1
Cebion, line 1: tantibus already has libspeex.so.1

image

Cebion (changes requested):

PR #1430, opened 2025-03-23
updated .json for CantoI

NotYourAveragePaladin:
I had updated the readme.md to reflect a R2R port, but forgot and missed the json. I apologize & this is going to fix it.


ports/cantoi/port.json
tabreturn, line 1: You still need to set:
"rtr": true,

tabreturn (changes requested):

kloptops (approved):

tabreturn (approved):

PR #1466, opened 2025-04-03
New Update for Beatblock 0.5.0 

Damian2809:
I hopefully resolved the problem with the file having a "different file ending" which was a error on the version before

### CFW Tests
- [ ] AmberELEC
- [ ] ArkOS
- [X] ROCKNIX
- [X] muOS

### Resolution Tests
Test all major resolutions:
- [ ] 480x320
- [X] 640x480
- [ ] 720x720 (RGB30)
- [X] Higher resolutions (e.g., 1280x720)

## File Structure
  - portname/
    - port.json
    - README.md
    - screenshot.jpg
    - cover.jpg
    - gameinfo.xml
    - Port Name.sh
    - portname/
      - <portfiles here>

## Additional Resources


ports/beatblock/Beatblock.sh
```diff
@@ -26,9 +26,10 @@ cd $GAMEDIR
 
 
 if [[ -f "gamedata/Beatblock" ]]; then
+   dos2unix ./patch/patch.diff
```
Cebion, line 29: no, why?

Cebion (changes requested):

ports/beatblock/Beatblock.sh
Damian2809: but i can remove it if you want, it just seems to resolve that error

Cebion: Then you need to convert the file and save it propely

Damian2809: Yeah they are and the last file was too but it still needed to be converted

Cebion: > Yeah they are and the last file was too but it still needed to be converted

That's not how this works! If the file needs to be converted it was not right in the first place.

Damian2809: > > Yeah they are and the last file was too but it still needed to be converted
> 
> That's not how this works! If the file needs to be converted it was not right in the first place. 

That file is directly grabbed from my X55, it has a LF ending. The Last was too and I still don't know why the last one didn't worked. So I just wanna make sure it works. I mean we use dos2unix on nearly all gmloader-next ports, while they also could just be already LF and they are but it is still being done

tabreturn (approved):

ben-willmore:
Since this seemed hard to resolve on Windows, I made a new PR to fix the CR/LF problems -- #1475 -- so am closing this one

PR #1473, opened 2025-04-05
iconoclasts: instructions to fix music issues

plaidman:
add instructions to fix music issues to iconoclasts described here: https://retrogamecorps.com/2021/11/27/guide-iconoclasts-on-retro-handhelds-via-portmaster/


ben-willmore:
Ideally we would add this to the port's .sh -- I can look into this

ben-willmore:
> Ideally we would add this to the port's .sh -- I can look into this

Not going to get to this for a while, so perhaps it's better just to go ahead.

Cebion (approved):

PR #1481, opened 2025-04-06
Add header to Wesnoth .gitignore

ben-willmore:
Add autogenerated header to Wesnoth .gitignore


tabreturn (approved):

tabreturn (changes requested):
Should this be placed in `ports/wesnoth/wesnoth/` instead?

kloptops:
> Should this be placed in `ports/wesnoth/wesnoth/` instead? Sorry, I'm not sure what this is actually for, but the location seems odd ...

No. This is one of the few files allowed in the ports top directory and is generally auto generated by the python script.

tabreturn (approved):

Cebion (approved):

PR #1487, opened 2025-04-08
Add pngquant to prepare_repo.sh

Doronimmo:
This should help make the repo less bloated


JeodC:
@kloptops Unsure about this, will defer to you.

ben-willmore:
This would require everyone who ever runs prepare_repo.py to have pngquant available, which doesn't seem reasonable to me.

Doronimmo:
might be better to add it here in release.yaml, looks like there's a PortMaster-New-Build-Release.sh but I can't find it

kloptops:
This will not work, for once pngquant will run on every release, this will break the checker/release scripts.

We are better off just doing a PR with all the cover/screenshot images pngquant'd.

Yes `PortMaster-New-Build-Release.sh` is on our server. I haven't made it public yet, but it is just a basic script.

Doronimmo:
Closing

PR #1488, opened 2025-04-08
Add --web argument in build_data.py for smaller chunks

Doronimmo:
chunk size is 25mb suitable for web upload (no need to clone)


kloptops (changes requested):
Did you copy the whole file to change one variable? 👀

Also how is this integrated into the build system?

You can already upload 25mb files named `<filename>.part.<XXX>`, just make sure you include an appropriately generated `.gitignore` file.

PR #1529, opened 2025-04-15
Update the readme for fheroes2 about the mouse cursor

chrisj951:
## Game Information
- **Title**: fheroes2
- **URL**:  https://portmaster.games/detail.html?name=fheroes2

README Only Update

Update the readme for fheroes2 about the mouse cursor
Indicate for some devices where the setting that must be enabled is

### DIFF
--- a/ports/fheroes2/README.md
+++ b/ports/fheroes2/README.md
@@ -45,6 +45,12 @@ DPAD LEFT - 4
 START - ENTER  
 SELECT -  ESC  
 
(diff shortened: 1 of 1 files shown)


ben-willmore (approved):
Ideally this would identify the platforms where this is needed, and the .sh would modify the cfg file as required. Nevertheless, this is useful information to have in the README.

Cebion (approved):

PR #1533, opened 2025-04-16
Fix missing .gitignore files.

kloptops:
Fixing some missing and or broken .gitignore files.


ports/openenroth/.gitignore
ben-willmore, line 1: This doesn't work on case insensitive file systems because it makes git ignore the entire openenroth directory.

kloptops: Blah, because this doesn't help the script.

Cebion (approved):

PR #1541, opened 2025-04-19
Start pruning xdelta3 duplicates

JeodC:
This pull request is a start for pruning unnecessary duplicate helper files. An xdelta3 binary is included with PortMaster GUI and does not need to be included with ports. This pull request changes 10 ports to remove their xdelta3 duplicate and instead reference the one included with PortMaster GUI.

Space-wise it doesn't make a massive difference yet, only saving about 2.63MB of space. A search found 62 more instances of xdelta3 duplicates though, which will save 18.93MB in the end. As we continue pruning duplicates, such as `gmtools` (oggenc, oggdec) and `7za`, as well as convert `legacy gmloader` to a runtime, we may end up saving our end users a couple hundred MBs.


Doronimmo:
You could add this to check if xdelta is available:
```
# check if we have new enough version of PortMaster that contains xdelta3
if [ ! -f "$controlfolder/xdelta3" ]; then
  echo "This port requires the latest PortMaster to run, please go to https://portmaster.games/ for more info." > /dev/tty0
  sleep 5
  exit 1
fi
```

kloptops (approved):

Cebion (approved):

PR #1554, opened 2025-04-22
Fixing Miss Driller air bottle graphic

designer2k2:
Hello,

this is a small fix for the air bottle graphic in Miss Driller. It removes the 1px "black square" around the air bottle. 

Comparing on the left the old file, the black square can be seen, on the right the new file with no square:
grafik

This is due to a dark blue border in the bmp file, i assume it is due to the conversion into the port. There the original pink was replaced by black.

Comparing the "original" before porting, the one in the port and my remake zoomed:
grafik

Link to the original air.bmp: https://github.com/yui0/missdriller/blob/master/data/bmp/air.bmp

Most visible is the difference when the air bottle is falling, but a video screen recording does not show it so well:
MissDrillerComp-ezgif com-optimize

Tested on the R36s running ArkOS 2.0 (File replaced manually on the SD card)

greetings!


cdeletre (approved):
You have an eye ! Thanks for this exhaustive PR. All looks good to me.

tabreturn (approved):

Cebion (approved):

PR #1571, opened 2025-04-27
Add GemRB runtime to JSON

tabreturn:
New selector runs on Pyxel.

### DIFF
--- a/ports/gemrb/port.json
+++ b/ports/gemrb/port.json
@@ -19,7 +19,7 @@
         ],
         "image": {},
         "rtr": false,
-        "runtime": null,
+        "runtime": "pyxel_2.2.8_python_3.11.squashfs",
         "reqs": [],
         "arch": [
             "aarch64"


ports/gemrb/port.json
```diff
@@ -19,10 +19,10 @@
         ],
         "image": {},
         "rtr": false,
-        "runtime": null,
+        "runtime": "pyxel_2.2.8_python_3.11.squashfs",
         "reqs": [],
```
kloptops, line 23: Put the arch under requirements so that it doesnt show up on x86_64 devices, then remove it from the "arch" requirements:

```suggestion
        "reqs": ["aarch64"],
```

The problem is if you specify a "runtime" it then ignores the "arch". 😄

kloptops (changes requested):

ports/gemrb/port.json
tabreturn: But I still keep the `"arch": [ "aarch64"]`?
https://github.com/PortsMaster/PortMaster-New/blob/2614c92ad4a460bf118ce80cdd968f01a4ece46d/ports/gemrb/port.json

kloptops (approved):

Cebion (approved):

PR #1579, opened 2025-04-28
Void Stranger: 1.1.2 not supported yet

gPhantasm:
This is just a small change to the instructions for Void Stranger's port, as it received an update today that is not yet supported by this port (the xdelta patch is only for v1.1.1, not v1.1.2). The instructions now clearly outline that 1.1.2 is not supported and that the supported version can be installed through Steam's open beta tab.

### DIFF
--- a/ports/voidstranger/README.md
+++ b/ports/voidstranger/README.md
@@ -2,15 +2,15 @@
 
 GitHub for this port- https://github.com/EzDzzIt/vsaarch64
  
-This port supports v1.1.1 of the Steam version of the game, or v1.1.1 of the itch.io version (itch version is patched to be equivalent to the Steam version).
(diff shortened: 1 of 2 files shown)


JeodC (approved):

gPhantasm:
> Looks like it's using the same runtime at least. Maybe can update it to not need an xdelta? There's also a newer gmloadernext version that could be used if the game has rumble support, and it doesn't seem to be using audio compression. I would test those new possibilities to see if an xdelta is necessary afterwards.

For sure. Part of updating the port to 1.1.2 will include testing if the fixes introduced in the xdelta are necessary or not. Void Stranger doesn't have rumble support, and the fixes in the xdelta were workarounds for the, at the time, experimental gmloader-next runtime being used. When I have the chance to work on updating it, I will try a newer runtime and see if everything works properly. I might also get in touch on the Discord server at that time to discuss compressing the audio as well.

kloptops (approved):

PR #1589, opened 2025-04-30
La-Mulana: Add compatibility with Steam files

ben-willmore:
Tested on Arkos, muOS.


ben-willmore:
Marking as draft -- forgot to update readme.

PR #1773, opened 2025-06-29
add new port Jagged Alliance 2

ztoth:
## Game Information
**New Port for**: Jagged Alliance 2
**URL**:
- https://ja2-stracciatella.github.io/
- https://www.gog.com/en/game/jagged_alliance_2

### CFW Tests
- [x] ArkOS
- [x] AmberELEC
- [x] ROCKNIX
- [x] muOS
- [ ] Knulli (Optional)
- [ ] Crossmix (Optional)
- [ ] Other (add here)

### Resolution Tests
Test all major resolutions:
- [ ] 480x320 (Optional)
- [x] 640x480
- [x] 720x720 (RGB30) (Optional)


ports/jaggedalliance2/Jagged Alliance 2.sh
```diff
@@ -0,0 +1,93 @@
...
+export SDL_GAMECONTROLLERCONFIG="$sdl_controllerconfig"
+
+# Ensure the conf directory exists
+mkdir -p $CONFDIR
+bind_directories ~/.ja2 "$CONFDIR"
+
+# Check if display meets the minimum resolution requirements
+echo "Display resolution is $DISPLAY_WIDTH x $DISPLAY_HEIGHT"
+if [ $DISPLAY_WIDTH -lt 640 ] || [ $DISPLAY_HEIGHT -lt 480 ]; then
+  echo "This game requires a minimum resolution of 640x480, exiting"
```
Cebion: We do not use echo for user messages.

pm_message "This port requires the latest PortMaster to run, please go to https://portmaster.games/ for more info."

pm_message inside funcs.txt explained

```
pm_message() {
    # Usage: pm_message "Some message here."
    if [[ "$CFW_NAME" == "muOS" ]]; then
        /opt/muos/extra/muxstart "$1"
    elif [[ "$CFW_NAME" == "ROCKNIX" ]]; then
        foot -F /bin/bash -c "echo \"$1\"; sleep 5" & 
    else
        echo "$1" > $CUR_TTY
    fi
}
```

echo for writing stuff into other files etc. is still okay.

ports/jaggedalliance2/port.json
```diff
@@ -0,0 +1,47 @@
...
+        "gameurl": "https://ja2-stracciatella.github.io/",
+        "developerurl": ""
+      }
+    ],
+    "availability": "paid",
+    "reqs": [
+      "power"
+    ],
+    "arch": [
+      "armhf",
```
Cebion: I don't see armhf and aarch64 binaries and a single libs folder.
So it's not armhf probably.

ztoth: My bad, I accidentally used the wrong folder which was used for armhf testing, fixed!

ports/jaggedalliance2/Jagged Alliance 2.sh
ztoth: I'm using echo because pm_message is currently broken on Rocknix, at least on latest stable. I'm getting this:
/roms/ports/PortMaster/funcs.txt: line 143: foot: command not found

ztoth: I can switch to pm_message for user facing messages, but we should look into what's happening with Rocknix. Maybe they switched to a different terminal emulator? We should consider changing the conditions in the pm_message implementation to check for what's available, instead of relying on CFW_NAME, wdyt?

tabreturn (approved):

PR #1780, opened 2025-07-01
Quick fix for Deltarune script to stop the deluge of help requests from Arkos users

Ganimoth:
quick temporary fix to get around problems with mounting the runtime on arkos. Tried by several people in the port help section and additionally tested by me on Arkos, muOS, Knulli, Rocknix and Amberelec on the LTS demo version from itch.

### DIFF
--- a/ports/deltarune/Deltarune.sh
+++ b/ports/deltarune/Deltarune.sh
@@ -26,19 +26,6 @@ cd $GAMEDIR
 # Permissions
 $ESUDO chmod +x $GAMEDIR/gmloadernext.aarch64
 
-# Mount the GMToolkit runtime
-TOOLKIT="$HOME/gmtoolkit"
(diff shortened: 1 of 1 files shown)


ports/deltarune/Deltarune.sh
```diff
@@ -26,6 +26,19 @@ cd $GAMEDIR
 # Permissions
 $ESUDO chmod +x $GAMEDIR/gmloadernext.aarch64
 
+# Mount the GMToolkit runtime
```
JeodC: This mounts the runtime each time the port opens, when it only needs to be mounted for patching. Just add `export ESUDO` before sourcing the patchscript.

Ganimoth: Thanks, thats a much better solution

Cebion (approved):

binarycounter (approved):

PR #1828, opened 2025-07-18
Update port.json

Fraxinus88:
To include "mono-6.12.0.122-aarch64.squashfs"

### DIFF
--- a/ports/dustaet/port.json
+++ b/ports/dustaet/port.json
@@ -22,7 +22,7 @@
     "image": null,
     "rtr": false,
     "exp": false,
-    "runtime": "mono-6.12.0.122-aarch64.squashfs",
+    "runtime": ["mono-6.12.0.122-aarch64.squashfs"],
     "store": [
       {
         "name": "Steam",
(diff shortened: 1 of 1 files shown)


ports/dustaet/port.json
```diff
@@ -22,7 +22,7 @@
     "image": null,
     "rtr": false,
     "exp": false,
-    "runtime": [],
+    "runtime": "mono-6.12.0.122-aarch64.squashfs",
```
kloptops:
```suggestion
    "runtime": ["mono-6.12.0.122-aarch64.squashfs"],
```

kloptops (changes requested):

ports/dustaet/port.json
binarycounter:
```suggestion
    "runtime": ["mono-6.12.0.122-aarch64.squashfs"],
```
Version 4 port files require "runtime" to be a list of strings instead of a single string.

binarycounter (changes requested):

ports/dustaet/port.json
binarycounter: @kloptops sniped me 😭

Fraxinus88: i suck at json :/

kloptops (approved):

PR #1837, opened 2025-07-22
Update port.json for steam games from v2 to v4

shafr:
This updates ports.json files from version v2 to version v4 for games that contain `steam links` (first batch to see if i'm moving in the right direction):
* adding corresponding images (either png or jpg)
* extracted store links to a store section. removed those from instructions.
* added paid or free
* add minlib_gc
* in one case added requitement for dual sticks.

**Checks from my side:**
* check schema compliance with v4 ( string to array changes for some values, null values, etc )
* check for pricing on the steam site via their API
* review of commit contents to check if everything looks ok. VS code does much better job of showing side-by-side comparison then github web.

If those would get approved - i'll go with gog and itch next.

I used python automation to update, sort fields, get steam data and validate schema. Also tried to use gemini-free api for comments update (links removal) with further manual review.

### DIFF
--- a/ports/annalynn/port.json
+++ b/ports/annalynn/port.json
@@ -1,5 +1,5 @@
 {
-  "version": 2,
+  "version": 4,
   "name": "annalynn.zip",
   "items": [
     "Annalynn.sh",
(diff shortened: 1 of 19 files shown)


tabreturn:
Thanks for this, @shafr. Before proceeding, I'd like to confirm that you've run `python3 tools/build_release.py --do-check` on your side. https://portmaster.games/packaging.html#creating-a-pull-request

kloptops:
As another note, the `"image"` attribute should in theory be `null`. Internally it gets overwritten. 😄

PR #1896, opened 2025-08-18
update Nazi Zombies port

ztoth:
This PR includes the following changes to the port:
- fix broken "invert look" functionality
- add support for devices without analog sticks, by making some changes in the default bindings during the first time run. This feature was requested by multiple players
- cleanup the README file to include the controls for stickless devices, as well as moving the "custom build" instructions section from the README into its dedicated src folder for better organization and a cleaner README file
- notify user to switch from libMali to panfrost on some devices, like the GKD Pixel 2, which runs a custom older version of Rocknix but the game only works with its panfrost driver


ztoth:
Hey folks! This could use some eyes. It has a couple of small but important changes to the port, requested by multiple users. We have tested them in the supported CFWs, see the relevant thread in the testing-n-dev channel

kloptops (approved):

PR #2028, opened 2025-10-04
Isles of Sea and Sky update 

Ganimoth:
Update to add zram setup for Arkos (thanks to bmdhacks) to prevent running out of memory during crashing. Retested on all firmwares including Knulli.


Dia2809 (approved):
Don't know if all arkOS devices support zram but looks good to me

Ganimoth:
> Don't know if all arkOS devices support zram but looks good to me

It worked for me on 353 image (353VS), 351MP image (regular 351MP and AeUX) and OGA image, so pretty much all proper arkos should be fine.

PR #2034, opened 2025-10-05
Fix Hexen 2 and Quakespasm not launching on Knulli

Slayer366:
## Game Information
- **Update Ports for**: Hexen 2 and Quakespasm
Adjust GL4ES config that caused Quakespasm and Hexen 2 to not launch on Knulli.  Skipping the two SDL_GL entries will forfeit the performance gain on Knulli that other CFWs will benefit from, but at least the games will work and aside from the beefy mods the base games and their original expansions will run just fine.

While it has been reported that it isn't launching on AmberElec's latest pre-release, I can't say for certain whether this is an issue with these two titles or if AmberElec's pre-release needs to have a few kinks worked out.

### CFW Tests
- [x] ArkOS
- [x] AmberELEC (Panda Conspiracy before newest pre-release)
- [x] Knulli (Optional)

### Resolution Tests
- [x] 480x320 (Optional)
- [x] 640x480

### DIFF
--- a/ports/hexen2/Hexen 2 - Portal of Praevus.sh
+++ b/ports/hexen2/Hexen 2 - Portal of Praevus.sh
@@ -27,9 +27,9 @@ else
 fi
 
 if [ "$LIBGL_FB" != "" ] && [ "${CFW_NAME^^}" != "KNULLI" ]; then
-  export SDL_VIDEO_GL_DRIVER="$GAMEDIR/gl4es.aarch64/libGL.so.1"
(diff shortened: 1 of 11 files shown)


Dia2809:
Why are you removing the LD_LIBRARY_PATH on all?

Slayer366:
> Why are you removing the LD_LIBRARY_PATH on all?

Kloptops suggested it because the libGL block already applies it.

Dia2809 (dismissed):

Slayer366:
Just realized I should really use the DEVICE_ARCH variable even though it will probably only ever be aarch64 from here on out, but I feel better having this done "right" so that in case another armhf situation lands that I won't have to come back and fix it later.

PR #2062, opened 2025-10-14
Update Thy Sword to gmloader-next

Fraxinus88:
Tested on relevant firmwares and resolutions

### DIFF
--- a/ports/thysword/Thy Sword.sh
+++ b/ports/thysword/Thy Sword.sh
@@ -74,7 +74,7 @@ fi
 # Swap buttons
 "$GAMEDIR/tools/SDL_swap_gpbuttons.py" -i "$SDL_GAMECONTROLLERCONFIG_FILE" -o "$GAMEDIR/gamecontrollerdb_swapped.txt" -l "$GAMEDIR/SDL_swap_gpbuttons.txt"
 export SDL_GAMECONTROLLERCONFIG_FILE="$GAMEDIR/gamecontrollerdb_swapped.txt"
-export SDL_GAMECONTROLLERCONFIG="`pm_message "$SDL_GAMECONTROLLERCONFIG" | "$GAMEDIR/tools/SDL_swap_gpbuttons.py" -l "$GAMEDIR/SDL_swap_gpbuttons.txt"`"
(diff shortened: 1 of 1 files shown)


ports/thysword/Thy Sword.sh
```diff
@@ -13,69 +13,78 @@ else
...
+	# Zip all game files into the game.port
+	zip -r -0 ./game.port ./assets/
+	rm -rf ./assets/
+fi
 
 # Swap buttons
 "$GAMEDIR/tools/SDL_swap_gpbuttons.py" -i "$SDL_GAMECONTROLLERCONFIG_FILE" -o "$GAMEDIR/gamecontrollerdb_swapped.txt" -l "$GAMEDIR/SDL_swap_gpbuttons.txt"
 export SDL_GAMECONTROLLERCONFIG_FILE="$GAMEDIR/gamecontrollerdb_swapped.txt"
-export SDL_GAMECONTROLLERCONFIG="`echo "$SDL_GAMECONTROLLERCONFIG" | "$GAMEDIR/tools/SDL_swap_gpbuttons.py" -l "$GAMEDIR/SDL_swap_gpbuttons.txt"`"
+export SDL_GAMECONTROLLERCONFIG="`pm_message "$SDL_GAMECONTROLLERCONFIG" | "$GAMEDIR/tools/SDL_swap_gpbuttons.py" -l "$GAMEDIR/SDL_swap_gpbuttons.txt"`"
```
Slayer366:
```suggestion
export SDL_GAMECONTROLLERCONFIG="`echo "$SDL_GAMECONTROLLERCONFIG" | "$GAMEDIR/tools/SDL_swap_gpbuttons.py" -l "$GAMEDIR/SDL_swap_gpbuttons.txt"`"
```

This is to set an environment variable and not to send a message to the screen for the user to see.  This needs to be echo as it's a command, in this instance, to pass an existing envar to the envar being set or altered, but pm_message will instead attempt to print the current envar to the screen instead and pm_message uses different commands depending on the platform to achieve this which can and will replace SDL_GAMECONTROLLERCONFIG with undesired text or even unset the envar completely.  pm_message is only to be used when sending a message to the user.

PR #2100, opened 2025-10-30
Update README.md and drift survivor

Fraxinus88:
Whoopsies


Dia2809 (approved):
Was that really a issue hehe, but sure 😃!

PR #2102, opened 2025-10-30
New Port for: Overbowed

tabreturn:
## Game Information
- **New Port for**: Overbowed
- **URL**: https://klardonics.itch.io/overbowed

### CFW Tests
- [x] ArkOS
- [x] AmberELEC
- [x] ROCKNIX
- [x] muOS
- [ ] Knulli (Optional)
- [ ] Crossmix (Optional)
- [ ] Other (add here)

### Resolution Tests
Test all major resolutions:
- [ ] 480x320 (Optional)
- [x] 640x480
- [x] 720x720 (RGB30) (Optional)

## File Structure
  - portname/
    - port.json
    - README.md
    - screenshot.png
    - cover.png
    - gameinfo.xml
    - Port Name.sh
    - portname/
      - <portfiles here>

## Additional Resources

### DIFF
--- a/ports/overbowed/Overbowed.sh
+++ b/ports/overbowed/Overbowed.sh
@@ -37,15 +37,15 @@ if [ -f "./assets/data.win" ]; then
   # prepare files
   mv "./assets/data.win" "./assets/game.droid"
   rm -f ./assets/*.exe ./assets/*.dll
-  # pick output name based on marker file
+  # pick wrapper based on marker file
   if [ -f "ver_itch" ]; then
-    out="./overbowed_itch.port"
+    wrapper="./overbowed_itch.port"
   elif [ -f "ver_steam" ]; then
-    out="./overbowed_steam.port"
+    wrapper="./overbowed_steam.port"
   fi
   # archive .port
   pm_message "Finalizing .port file ..."
-  zip -r -0 "$out" "./assets/"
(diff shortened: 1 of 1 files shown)


ports/overbowed/Overbowed.sh
```diff
@@ -0,0 +1,67 @@
...
+# cd and set permissions
+cd $GAMEDIR
+> "$GAMEDIR/log.txt" && exec > >(tee "$GAMEDIR/log.txt") 2>&1
+$ESUDO chmod +x $GAMEDIR/gmloadernext.aarch64
+
+# prepare game files
+if [ -f "./assets/data.win" ]; then
+  pm_message "Preparing game files ..."
+  # determine version
+  checksum=$(md5sum "assets/data.win" | awk '{print $1}')
```
Dia2809, line 31: This can make issues if the game is updated, like not much just a simple little update and it will now try the steam version instead of the itch io one?

So maybe instead of that you could also check the steam version so it matches? and then do a else doing a pm_message " Uncompatible Game version"

tabreturn: Thanks, but this is necessary because the itch.io version runs on an older version of the wrapper. If anything, if it's updated (which is unlikely given how long it hasn't been), it'll match the Steam wrapper. Hence, defaulting to 'steam' makes sense in any scenario where the current itch.io checksum fails.

PR #2113, opened 2025-11-02
Update Q2Pro - add x86_64 support

Slayer366:
## Game Information
- **Update Port for**: Q2Pro
Add x86_64 support
Update shell scripts and port.json to reflect x86_64 addition
Uses squoosh to reduce file size of cover and screenshot
Add license files for additional libs needed by RetroDeck
Add gptokeyb config for devices with naturally placed triggers
Add placeholder video folders for .cin cinematic files for Quake II and its expansions

### CFW Tests
- [x] ArkOS
- [x] Other (RetroDeck)

### Resolution Tests
- [x] 480x320 (Optional)


ports/q2pro/Q2Pro.sh
```diff
@@ -49,12 +49,23 @@ if [ ! -f "firsttime" ]; then
...
+elif [[ "${DEVICE_ARCH}" == 'x86_64' ]] || [[ "${CFW_NAME^^}" == "RETRODECK" ]]; then
+    # x86_64
+    BINARY=q2pro_x86_64
 else
+    # aarch64
     BINARY=q2pro_glsl
 fi
 
-$GPTOKEYB "$BINARY" -c "$GAMEDIR/q2pro.gptk" &
+if [[ "${DEVICE_NAME^^}" == "X55" ]] || [[ "${DEVICE_NAME^^}" == "RG353P" ]] || [[ "${DEVICE_NAME^^}" == "RG40XX-H" ]] || [[ "${CFW_NAME^^}" == "RETRODECK" ]]; then
```
Dia2809, line 62: Why do those get a special gptk?

Slayer366: If you've seen my other ports, you'd know that I do this for each of mine that I have swapped the triggers and bumpers for devices that have naturally placed triggers such as those above.  These aren't the only ones, but they're the ones I've added support for.  The first device I'd owned that had naturally placed triggers was the RG353P and later the RG40XX-H.  I'd added the X55 since Cebion was testing with that one quite some time back and he could let me know if the detection in this check worked for his unit.  He didn't really mind if the fire button was on the bumper as long as the mapped button was in close proximity, but I do care that if we have triggers, set them up just like they'd be on a Playstation or Xbox controller.  I'd suggested this to EvilJagaGenius was he was porting this one, but I think it just got missed when troubleshooting the troublesome graphical issues with Rocknix at the time.

PR #2130, opened 2025-11-07
STEAMWORLD DIG/HEIST added 32bit export

NotYourAveragePaladin:
Just needed the 'export 32_BIT="Y" 'line added to the .sh's of these and audio is working now on MuOS. Steamworld Dig 2 does not have this export but audio seems to work still for some reason.


Dia2809 (approved):
can you please name your PR something actual useful, like what game too??? It's a PITA sometimes to find the right PR

PR #2131, opened 2025-11-08
Add build script for creating runtime zip archives

kloptops:
Introduces `tools/build_runtimes_zips.py` to automate the creation of runtime packs for offline users.

- Generates `full` and `popular` runtime archives for each architecture.
- Automatically splits the `full` archive into smaller parts when it exceeds GitHub's asset size limits.
- Calculates runtime popularity to create an optimized, smaller pack.
- Creates a `runtimes_zips.json` manifest for PortMaster to consume.
- Fixed a few fucking .gitignore files.


kloptops:
> Should those .gitignore be in there?

Who else is gonna update them, they won't cause the ports to be updated. 😄

PR #2183, opened 2025-11-25
Fix Max Payne port on platforms with non standard unzip

orktes:
At least on Rocknix it seems unzip doesn't expand globs. This change expands globs already in bash so unzip wont have to.

https://github.com/orktes/max_arm64/compare/v0.1.40...v0.1.42


ports/maxpayne/maxpayne/launcher.love
Cebion, line 1: What was changed on this file?

orktes: @Cebion nothing. Sadly something just changes inside archive. Very likely created at timestamps etc for the files. Will look into how I could make the archive remain the same if nothing in the launcher changes.

orktes: Diff between the tags is here https://github.com/orktes/max_arm64/compare/v0.1.40...v0.1.41

Cebion, line 1: In Git you can control what files get submitted for change. So if you only want to touch the pathcscript and version.txt that's all you commit. No other files should be touched.

orktes: @Cebion PR is created automatically based on the automated CI/CD builds. I'm only inputting description manually.

orktes: This way source of truth really is whats in the max_amd64 repo and its releases and not me trying (and failing) manually to update right files.

orktes: https://github.com/orktes/max_arm64/commit/2fcdf1bac22b054ad4591eaf1a8d61739314bb7c should make the love archive now stable cross builds. I can tag a new build so this point onwards its locked down.

Cebion: > @Cebion PR is created automatically based on the automated CI/CD builds. I'm only inputting description manually.

Just me being overly cautious because people try to smuggle stuff in or just update binaries for the sake of it and then break it for all even though they just wanted to update a single line in a text file :P

orktes: @Cebion totally get it and I've been just lazy it fixing the archive building. IMO full benefit of the my build pipeline also comes when people can expect its to be exactly what they'd get if building it from the repo. So completely accepting the call out on this.

PR #2223, opened 2026-01-01
Isles of Sea and Sky - updated patchscript and specified supported Steam version

Ganimoth:
The update process currently wont work on latest version (data.win wont even load on using UTMT nightly), so until it gets somewhat sorted I have to revert to older manifest. That cant be done on Itch.io so, so support for that has to be removed, at least for a while. Also the patchscript is revised in a way that lowers RAM usage during data.win patching and texture compression. Retested on a couple units, including 1GB Arkos device (usually the most RAM bottlenecked)


Dia2809 (dismissed):

Ganimoth:
Actually, hold on, looks like Itch.io version has not been updated yet, I need to rebuy it and confirm. Making this a draft for now

PR #2225, opened 2026-01-01
Update 3 ports' swap creation to work with dArkOS

Slayer366:
## Game Information
- **Update Port for**: Duke Nukem 3D: Alien World Order
- **Update Port for**: Ion Fury
- **Update Port for**: Rise of the Triad

eduke32 absolutely will not launch without an existing swap.
Added dArkOS detection and create swap using zram for dArkOS (swapfile is not possible with btrfs & kernel 4.19)
Re-tested on both ArkOS and dArkOS to ensure proper swap gets created per CFW after swap block addition.

### CFW Tests
- [x] ArkOS
- [x] dArkOS

NOTE: Open Transport Tycoon Deluxe may or may not require the same update to its shell script.

### DIFF
--- a/ports/duke3dawo/Duke3D - Alien World Order.sh
+++ b/ports/duke3dawo/Duke3D - Alien World Order.sh
@@ -45,13 +45,23 @@ export SDL_GAMECONTROLLERCONFIG="$sdl_controllerconfig"
 # Ensure Swap space is prepared or eduke32 may crash or fail to launch
 if [[ $CFW_NAME == *"ArkOS"* ]] || [[ $CFW_NAME == *"ODROID"* ]]; then
     pm_message "Preparing Swap File, please wait..."
-    [ -f /swapfile ] && $ESUDO swapoff -v /swapfile
-    [ -f /swapfile ] && $ESUDO rm -f /swapfile
-    $ESUDO fallocate -l 384M /swapfile
-    $ESUDO chmod 600 /swapfile
-    $ESUDO mkswap /swapfile
-    $ESUDO swapon /swapfile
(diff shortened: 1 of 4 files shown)


ports/duke3dawo/Duke3D - Alien World Order.sh
```diff
@@ -45,13 +45,23 @@ export SDL_GAMECONTROLLERCONFIG="$sdl_controllerconfig"
...
+    if [[ $CFW_NAME == "dArkOS" ]]; then
+        [ -e /dev/zram0 ] && $ESUDO swapoff -a
+        [ -e /dev/zram0 ] && $ESUDO zramctl --reset /dev/zram0
+        [ -e /dev/zram1 ] && $ESUDO zramctl --reset /dev/zram1
+        [ -e /dev/zram2 ] && $ESUDO zramctl --reset /dev/zram2
+        modprobe zram
+        $ESUDO zramctl --find --size 384M
+        $ESUDO mkswap /dev/zram0
+        $ESUDO swapon /dev/zram0
+    else
```
Dia2809, line 57: Wont this make it so that it runs a swapfile on all cfws, doesnt matter if it can do that or not or not needed etc?

Slayer366: No... where do you see that?

Slayer366: You're not reading the whole if block.
if [[ $CFW_NAME == \*"ArkOS"\* ]] || [[ $CFW_NAME == \*"ODROID"\* ]]; then
       if [[ $CFW_NAME == "dArkOS" ]]; then
       else
       fi
The 'else' here will only execute if it's any other ArkOS flavor or that 'ODROID' and only if it isn't dArkOS.
It's referred as a nested if

PR #2281, opened 2026-02-04
Fix controller for independent 2 player splitscreen

jrosetto:
Fix controller for independent 2 player splitscreen

### DIFF
--- a/ports/openmrac/openmrac/conf/Franticware/openmrac/settings-sdl2.dat
+++ b/ports/openmrac/openmrac/conf/Franticware/openmrac/settings-sdl2.dat
@@ -1,0 +1,42 @@
+renderer 1        // 0 - GL Compatibility, 1 - GL Core 3, 2 - GLES2 
+fullscreen 0        // 0 - windowed, 1 - fullscreen 
+screen_x 1280        // screen resolution 
+screen_y 720 
+vsync 1        // 0 - vsync off, 1 - vsync on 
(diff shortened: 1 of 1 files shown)


ports/openmrac/openmrac/openmrac.gptk
Cebion, line 1: Did you actually test this? Because this breaks the controls for this port for all players.

gptokeyb stands for gamepad to keyoard translation.

So the a button corresponds to the up arrow on the keyboard, so accelerating in this instance.

a=a
b=b

would set the a button to the a key on the keyboard.

jrosetto: Yes, and just tested again to confirm...

Portmaster version has left and right analog tied to the same buttons making it impossible to drive either car at the same time.  My change makes it so player 1 gets the left side (left analog and dpad) and player 2 gets the right side (a,b,x,y and right analog).

The game only maps player 1 on first launch.  End user has to go into controls and map player 2.

I have added the settings file so the controls are mapped and the end user has to do nothing.

kloptops (approved):

ports/openmrac/openmrac/openmrac.gptk
jrosetto: Any reason this isn't moving?

This is a simple modification.  As long as the keys are bound to my modification it allows two players to play on one device without interference.

PR #1469, opened 2025-04-04
Optimize all port.json files.

bmdhacks:
This PR is more of a RFC but I'm just posting it here to continue the conversation.

Problem: With the growth of PortMaster, the lack of port database qua…lity is providing friction for surfacing good ports to users.

Solution: This is the first step in cleaning up the port information.  A new port.json schema is defined, v5, and all ports have been migrated to this format.  Included in this format are links to itch.io, IGDB, and Steam so that we can further correlate game popularity with their available ports.  Two scripts are add that use the IGDB information to catalog game popularity metrics, and view this data.

PR #2286, opened 2026-02-08
Add missing store link metadata

offalynne:
Store links for all non-RTR ports

### DIFF
--- a/ports/absolutereflex/port.json
+++ b/ports/absolutereflex/port.json
@@ -21,6 +21,13 @@
     "exp": false,
     "runtime": null,
     "reqs": [],
+    "store": [
+      {
+        "name": "Itch.io",
+        "gameurl": "https://jukio.itch.io/absolute-reflex",
+        "developerurl": "https://itch.io/profile/jukio"
+      }
+    ],
     "arch": [
       "aarch64"
     ]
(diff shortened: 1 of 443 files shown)


tabreturn (approved):

offalynne:
@kloptops

offalynne:
Is it possible to have maintainer assistance to get this merge done?

I am deeply frustrated by the low quality of metadata which makes it more difficult to externally find & purchase the games required to use PortMaster than it reasonably should be.

Moreso when realizing this has already been attempted and both efforts are apparently gone to waste https://github.com/PortsMaster/PortMaster-New/pull/1469

As further changes are made conflicts pile up and my confidence this will ever actually be addressed diminishes.

bmdhacks:
We've explained the blocker: ports are rebuilt and redownloaded on metadata changes, possibly blowing up our quota. You are not entitled to other peoples time in open source unfortunately, so your options are:

1. Hire somebody to fix the issue of port.zip tied to metadata.
2. Do a one time donation to portmaster to upgrade our serving quotas and take the hit of committing this. (I would prefer we commit my version though)
3. Do the coding work yourself to fix the blocker.

We can all sit around and complain that too many people submit ports and nobody works on PM infrastructire besides klops, but the only real lever you have is to fix it yourself in some fashion.

offalynne:
> _I would prefer we commit my version_

👍

PR #2373, opened 2026-04-01
feat: new port wizball

craigchandler:
## Game Information
- **New Port for**: Wizball
- **URL**: https://github.com/craigchandler/wizball-remake

### CFW Tests
- [X] ArkOS
- [X] AmberELEC
- [X] ROCKNIX
- [X] muOS
- [X] Knulli (Optional)
- [X] Crossmix (Optional)
- [X] Other (add here)

### Resolution Tests
Test all major resolutions:
- [ ] 480x320 (Optional)
- [X] 640x480
- [X] 720x720 (RGB30) (Optional)
- [X] Higher resolutions (e.g., 1280x720)

## File Structure
  - portname/
    - port.json
    - README.md
    - screenshot.png
    - cover.png
    - gameinfo.xml
    - Port Name.sh
    - portname/
      - <portfiles here>

## Additional Resources


Cebion (commented):
Any reason why thee is a armhf version of this port supplied?

ports/wizball/port.json
tabreturn, line 1: This is an old version (where the current version is 4). You can use the JSON creator to generate this -- https://portmaster.games/port-json.html

ports/wizball/WizBall.sh
```diff
@@ -0,0 +1,59 @@
...
+# If $directory is empty (some CFWs don't set it), derive GAMEDIR from
+# the script's own location: script lives at ports/WizBall.sh so the
+# port folder is the wizball/ sibling directory.
+if [ ! -d "$GAMEDIR" ]; then
+  GAMEDIR="$(dirname "$(readlink -f "$0")")/wizball"
+fi
+cd "$GAMEDIR" || { echo "FATAL: cannot cd to $GAMEDIR" > /tmp/wizball_fatal.txt; exit 1; }
+
+# Simple redirect to logfile - avoid process substitution (unreliable on embedded shells)
+exec > "$GAMEDIR/log.txt" 2>&1
```
tabreturn:
```suggestion
> "$GAMEDIR/log.txt" && exec > >(tee "$GAMEDIR/log.txt") 2>&1
```

tabreturn (changes requested):

ports/wizball/port.json
craigchandler: Updated the port metadata to v4 and corrected the package name to wizball.zip. The old WizBall.zip artifact has been removed as well. Ready for re-review.

ports/wizball/WizBall.sh
```diff
@@ -0,0 +1,59 @@
...
+get_controls
+
+
+GAMEDIR="/$directory/ports/wizball"
+# If $directory is empty (some CFWs don't set it), derive GAMEDIR from
+# the script's own location: script lives at ports/WizBall.sh so the
+# port folder is the wizball/ sibling directory.
+if [ ! -d "$GAMEDIR" ]; then
+  GAMEDIR="$(dirname "$(readlink -f "$0")")/wizball"
+fi
```
Slayer366, line 26: I see exactly what you have going on here and it's not a bad idea for a fallback.  I would like to point out that while this can work if one tries to run it without relying on PortMaster's control.txt, this won't work on systems running muOS or RetroDeck (I know you don't have x86_64 in here, but just an example) as the shell scripts and the game folders are segregated on these CFWs and the control.txt rarely fails FWIW.  I won't say you should remove this, but if control.txt fails, then even DEVICE_ARCH will be empty for the binary as well among other envars.

ports/wizball/WizBall.sh
```diff
@@ -0,0 +1,59 @@
...
+echo "DEVICE_ARCH=${DEVICE_ARCH}"
+
+# SDL controller mapping provided by PortMaster
+export SDL_GAMECONTROLLERCONFIG="$sdl_controllerconfig"
+
+# If you ever ship extra libs, uncomment this and use libs.aarch64 / libs.armhf
+# export LD_LIBRARY_PATH="$GAMEDIR/libs.${DEVICE_ARCH}:$LD_LIBRARY_PATH"
+
+BIN="wizball.${DEVICE_ARCH}"
+echo "BIN=$BIN  exists=$(test -f "$BIN" && echo yes || echo NO)"
```
Slayer366:
```suggestion
echo "BIN=$BIN
```
I'm suggesting this because BIN will never _not_ exist since you solely declared it directly above.  The only thing that can go wrong is if DEVICE_ARCH doesn't return anything (which honestly would indicate a bigger problem) in which BIN would then have "wizball." meaning it would still exist as an envar with an entry even if DEVICE_ARCH is somehow empty.

ports/wizball/WizBall.sh
```diff
@@ -0,0 +1,59 @@
...
+
+# PortMaster helper (sets up env / permissions / etc.)
+pm_platform_helper "$BIN"
```
Slayer366:
```suggestion
pm_platform_helper "$GAMEDIR/$BIN"
```
pm_platform_helper suggests using the full path to the binary.

ports/wizball/WizBall.sh
```diff
@@ -0,0 +1,59 @@
...
+cd $GAMEDIR
+chmod +x "$BIN"
+
+# Set renderer backend to GLES2 - this is the most compatible option across all devices, and the fallback if Vulkan support is spotty.
+WIZBALL_RENDERER_BACKEND=gles2
+
+# Start keymapper for this binary (kill-mode hotkey support)
+export LD_LIBRARY_PATH="$controlfolder:$LD_LIBRARY_PATH"
+$GPTOKEYB "$BIN" -c "./wizball.gptk" &
+SDL_GAMECONTROLLERCONFIG="$sdl_controllerconfig" ./$BIN
```
Slayer366:
```suggestion
./$BIN
```
While harmless, you already have this envar exported above making this redundant.

Slayer366 (changes requested):

ports/wizball/WizBall.sh
craigchandler: applied, thanks

ports/wizball/WizBall.sh
```diff
@@ -0,0 +1,59 @@
...
+pm_platform_helper "$BIN"
+cd $GAMEDIR
+chmod +x "$BIN"
+
+# Set renderer backend to GLES2 - this is the most compatible option across all devices, and the fallback if Vulkan support is spotty.
+WIZBALL_RENDERER_BACKEND=gles2
+
+# Start keymapper for this binary (kill-mode hotkey support)
+export LD_LIBRARY_PATH="$controlfolder:$LD_LIBRARY_PATH"
+$GPTOKEYB "$BIN" -c "./wizball.gptk" &
```
tabreturn: Looks like you're using `$GPTOKEYB2` (not `$GPTOKEYB`). Best to pick one or the other. Typically, if you're opting for `$GPTOKEYB2`, you'd give the target file a `.ini` extension.

ports/wizball/WizBall.sh
tabreturn, line 1: For executing the 32-bit binary, you'd include a (conditional) `export PORT_32BIT="Y"` -- for example: https://github.com/PortsMaster/PortMaster-New/blob/08bfa422a057459af3b5989271ecaeed14b34e01/ports/happyhills/The%20Happyhills%20Homicide.sh#L16

Honestly, I'm not sure it's worth complicating things by including a 32-bit binary. All PortMaster handhelds support 64-bit (it's just that some can't handle 32-bit) -- so 64-bit is always preferred, anyhow.

tabreturn (changes requested):

ports/wizball/WizBall.sh
craigchandler: removed 32bit binary

ports/wizball/WizBall.sh
craigchandler: renamed to ini and switched to $GPTOKEYB2

ports/wizball/gameinfo.xml
```diff
@@ -0,0 +1,13 @@
+<?xml version='1.0' encoding='utf-8'?>
+<gameList>
+    <game>
+        <path>./Wizball.sh</path>
```
Slayer366:
```suggestion
        <path>./WizBall.sh</path>
```
Case sensitive

ports/wizball/cover.png
Slayer366, line 1: Drop this and screenshot.png into https://squoosh.app/
Select OxiPNG for the output
Tick the box for "Reduce Palette"
Click the blue download button once the end filesize is small enough (under 100KB) and re-upload them.
That should reduce the capacity of both of these from 270KB each to < 100KB each with no noticeable quality loss.

ports/wizball/WizBall.sh
Slayer366: For some reason GH decided to ignore the change.

Slayer366 (changes requested):

ports/wizball/cover.png
craigchandler: Done, thanks

Slayer366 (approved):

tabreturn (approved):

PR #2455, opened 2026-05-16
Savingprincessnext

Fraxinus88:
Tested on relevant CFWs and resolutions. Update to aarch64 gmloader.next


ports/savingprincess/port.json
tabreturn, line 1: Are inst/Inst_md meant to be identical? Like, usually one would leave the _md entries as null (unless these had additional formatting)

Fraxinus88: Okies have removed!

tabreturn (approved):

kloptops (approved):

PR #2529, opened 2026-06-29
Fix incorrect match on undertaleredyellow mod zip

alydevs:
Matching on UNDERTALE.exe returns a false positive if a user has copied the base game files to their system as a zip alongside the mod zip. This change searches for UTRY.*xdelta, to match a file that should only ever be in the UTRY zip, ensuring that it grabs the correct match even if there are other zips in the assets folder for some reason. 

## Game Information
- **Bug fix for**: Undertale Red and Yellow

### DIFF
--- a/ports/undertaleredyellow/undertaleredyellow/tools/patchscript
+++ b/ports/undertaleredyellow/undertaleredyellow/tools/patchscript
@@ -81,11 +81,11 @@ fi
 
 # First we need to verify the user placed everything correctly.
 prepare_assets() {
-    # Identify ut-red-yellow.zip
+    # Identify ut-red-and-yellow.zip
     echo "Looking for Undertale Red & Yellow zip file..."
     sleep 1
     for zipfile in "$DATADIR"/*.zip; do
-        if unzip -l "$zipfile" 2>/dev/null | grep -q "UNDERTALE\.exe"; then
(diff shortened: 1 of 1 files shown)


Producdevity (approved):

kloptops:
Unfortunately we have removed this port due to it no longer being actively maintained, it is open to being fully resubmitted.

alydevs:
What does "actively maintained" mean in this context? Surely receiving a PR would mean a package is "maintained". This package works fine if the right things are done in the right order, and this PR adds a fix for an edge case that the original author didn't consider. No reason to remove working features if they're entirely isolated to one package.

kloptops:
The original porter doesn't plan on updating it, and has requested it was removed.

JeodC:
> What does "actively maintained" mean in this context? Surely receiving a PR would mean a package is "maintained". This package works fine if the right things are done in the right order, and this PR adds a fix for an edge case that the original author didn't consider. No reason to remove working features if they're entirely isolated to one package.

Hello, I am the original author, and I do [actively maintain](https://github.com/JeodC/RHH-Ports/tree/main/ports/released/gamemakerengine/utyellow) it externally. I saw your pull request here last week and checked: it looks like for my current version the false positive you describe is no longer possible. If you find any further improvements can be made, I welcome collaboration! Sorry for the confusion.

alydevs:
Thank you for the reply!

PR #2587, opened 2026-07-27
Deltarune update

Ganimoth:
Few tweaks to partially aleviate a memory leak at the final boss in the ch5 + improve performance for  said fight


JanTrueno (commented):
Nice, could you perhaps add some more info on which versions of the game work right now? I think its outdated.

Ganimoth:
> Nice, could you perhaps add some more info on which versions of the game work right now? I think its outdated.

It should still work with the latest version, though I havnt played through the whole game.

JanTrueno (approved):

Cebion (approved):

PR #2592, opened 2026-07-29
dARKOS Flashing Screen Fix

jrosetto:
pm_message causes dARKOS screen to flash between the launcher and the portmaster message window.  Removing the "Starting Launcher" fixes the issue completely.

### DIFF
--- a/ports/maxpayne/Max Payne.sh
+++ b/ports/maxpayne/Max Payne.sh
@@ -42,7 +42,6 @@ export SDL_GAMECONTROLLERCONFIG="$sdl_controllerconfig"
 
 source $controlfolder/runtimes/"love_11.5"/love.txt
 
-echo "Starting Launcher"
 
 # Remove debug log
 rm -f "debug.log"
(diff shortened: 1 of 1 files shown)


ports/maxpayne/Max Payne.sh
```diff
@@ -42,7 +42,7 @@ export SDL_GAMECONTROLLERCONFIG="$sdl_controllerconfig"
 
 source $controlfolder/runtimes/"love_11.5"/love.txt
 
-pm_message "Starting Launcher"
```
bmdhacks, line 45: Thanks for the fix! Just change it to an `echo` don't comment it out.

bmdhacks (changes requested):

ports/maxpayne/Max Payne.sh
jrosetto: Why didn't I think of that 😄

Dia2809:
even smarter, lets remove that line entirely?

ports/maxpayne/Max Payne.sh
```diff
@@ -42,7 +42,7 @@ export SDL_GAMECONTROLLERCONFIG="$sdl_controllerconfig"
...
+echo "Starting Launcher"
```
Cebion:
```suggestion
```

PR #2593, opened 2026-07-30
Update Open Chaos aarch64 build and libraries

jckhng:
- Rebuild the Open Chaos executable and bundled libraries against Ubuntu
  20.04/Focal.
- Lower requirements from GLIBC 2.38 / GLIBCXX 3.4.31 to GLIBC 2.29 /
  GLIBCXX 3.4.26.
- Include the explosion particle bounds fix from UltimaBeaR/OpenChaos#7.
- Replace OpenAL Soft 1.25.1 with 1.24.3.


ports/openchaos/openchaos/libs.aarch64/libatopology.so.2
bmdhacks, line 1: Why do you have two copies of each library?

ports/openchaos/openchaos/libs.aarch64/libSDL3.so.0
bmdhacks, line 1: Which version of libSDL3 is this? you can't ship stock SDL3 on ports or itll break system SDL's which sometimes do screen rotation.

ports/openchaos/openchaos/libs.aarch64/libatopology.so.2
jckhng: ive clean up. think it had some issue with symlink earlier. but it is resolved now in the latest commit. thank you for pointing it out!

ports/openchaos/openchaos/libs.aarch64/libSDL3.so.0
jckhng: its built from your sdl shim at https://github.com/bmdhacks/SDL/tree/sdl2-backend. thanks!

ports/openchaos/openchaos/licenses/SDL3-SHIM-PROVENANCE.txt
bmdhacks, line 1: This is a bit reactionary to my earlier question, I don't think a whole file is necessary

ports/openchaos/openchaos/assets/.gitkeep
bmdhacks, line 1: Did you intend to delete this? This will prevent this directory from being present for users to copy assets into.

ports/openchaos/README.md
```diff
@@ -6,6 +6,58 @@ The current release package is `aarch64` only. It ships an SDL3 shim that uses
 the device's SDL2 runtime, plus the required ARM64 shared libraries under
 `openchaos/libs.aarch64/`.
 
+The SDL3 shim is built from bmdhacks' `sdl2-backend` branch at the pinned
```
bmdhacks, line 9: This is also a bit reactionary.  Can you read AGENTS.md in the base repo to clarify to your agent that this PR is bordering on rejection?  We'll keep it open because I appreciate your efforts, but let's get it back into a reasonable shape here.

Cebion:
Closed due to inactivity

PR #2602, opened 2026-08-02
Update Balatro with handheld layout and performance options

Guandor:
## Summary

- bring the improvements from [Guandor/Balatro-Lite](https://github.com/Guandor/Balatro-Lite) into the existing `balatro` package, launcher, and catalogue identity so this updates the current port instead of creating a separate entry
- add an optional small-screen layout, optional performance changes, and guided controller mapping
- rebuild the patched game safely when the selected options or bundled patches change, while retaining the purchased donor file
- update the installation documentation, 640x480 gameplay screenshot, metadata, and license layout

## Validation

- `python tools/build_release.py --do-check` — 1 updated, 0 broken
- `python tools/build_release.py --quick-build balatro`
- `bash -n ports/balatro/Balatro.sh`
- `git diff --check`
- inspected the generated `balatro.zip`: 30 entries and no `.exe`, `.love`, or `.7z` donor files

## Device testing

Device testing is complete across the required PortMaster device/CFW matrix.


ports/balatro/Balatro.sh
JanTrueno, line 1: Although i'm sure this works. For cleanliness please use the patcher program with its own patching bash script.

JanTrueno, line 1: Since you are updating the port. It would be nice to use the Love2D runtime instead of using binaries in the portfolder.

ports/balatro/balatro/patches/small_screen.lua
JanTrueno, line 1: Are these patches made from the ground up? Or they based on Balatro code?

JanTrueno (changes requested):

ports/balatro/Balatro.sh
Guandor: Thanks for the feedback, fixed now.

Guandor: Thanks for the feedback, fixed now. It now uses the patcher but with some custom functionality as well for button mapping and patching customizations. (small screen layout vs original, performance improvements on vs off)

ports/balatro/balatro/patches/small_screen.lua
Guandor: I'm not sure if I understand the question correctly but if you ask whether it contains original game code or not, most of the added logic was written for this port, but some UI overrides are based on Balatro’s existing functions and the patchscript contains small source fragments used to locate and modify the player’s copy.

ports/balatro/Balatro.sh
JanTrueno: This part looks good, thank you.

JanTrueno: Is there not a possibility to keep all the patches contained in the patchscript? It might just be me, but this doesnt seem organised?

ports/balatro/Balatro.sh
```diff
@@ -12,120 +12,265 @@ else
...
-  sed -i 's/self.F_HIDE_BG = false/self.F_HIDE_BG = true/g' globals.lua
+# Set to 1 to draw a small readout in the top-left corner: frame rate, Lua heap
+# size, and how many live objects each frame has to walk. Worth turning on if
+# the game slows down over a long run -- a climbing heap and a climbing object
+# count have different causes and different fixes.
+PERF_HUD=0
+export BALATRO_PM_PERF_HUD="$PERF_HUD"
 
-  # change controller mapping (swap A/B,X/Y to match the physical buttons) for TSP/BRICK
-  if [ "${DEVICE_NAME}" = "TrimUI Smart Pro" ] || [ "${DEVICE_NAME}" = "TrimUI Brick" ]; then
```
Dia2809, line 59: Why is this just not used?

ports/balatro/Balatro.sh
```diff
@@ -12,120 +12,265 @@ else
...
+  GAMEFILE="Balatro.love"
 elif [ -f "balatro.love" ]; then
-    GAMEFILE="balatro.love"
+  GAMEFILE="balatro.love"
 fi
 
-if [ -f "$GAMEFILE" ]; then
-  # Extract globals.lua
-  ./bin/7za.${DEVICE_ARCH} x "$GAMEFILE" globals.lua
+# The layout and the performance changes are the player's to choose, and both
```
JanTrueno: Please keep the comments to a minimum.

JanTrueno (dismissed):

ports/balatro/balatro/patches/small_screen.lua
JanTrueno: yeah i mean that these cannot be full lua scripts from balatro with some edits made into.

ports/balatro/port.json
```diff
@@ -9,21 +9,27 @@
...
         "genres": [
             "casino/card",
             "puzzle"
         ],
         "image": {},
         "rtr": false,
-        "runtime": null,
+        "runtime": [
+            "love_11.5"
+        ],
```
Dia2809: There is no Love runtime, its always installed via PM because PM itself uses for its startup Animation. Its just located in the runtimes folder

```suggestion
        "runtime": null,
```

ports/balatro/port.json
Dia2809, line 1: People will be confused when you constantly talk about this being a updated version. Just leave it as it is and put a little note into the readme

ports/balatro/balatro/patches/small_screen.lua
Guandor: They're not :)

ports/balatro/Balatro.sh
Guandor: It's no longer required with the new button mapping utility. The previous solution covered only Trimui Brick, but I faced the same issue on my RG40XXV, so I figured I should make a device-agnostic solution.

ports/balatro/Balatro.sh
Guandor: Just to clarify, are you referring to the build/rebuild logic in Balatro.sh, or to the separate Lua modules under balatro/patches? The launcher currently handles setup, determines whether the patched archive is stale, invokes PortMaster’s patcher, and validates the result. The actual modifications to the player’s game are performed by tools/patchscript and the Lua modules it installs. I kept balatro/patches separate because patchscript handles the archive patching process, while the Lua files are modules inserted into the patched game. They are quite long scripts, so makes it easier to maintain and keep separation of concerns.

ports/balatro/Balatro.sh
Guandor: Removed all comments

ports/balatro/port.json
Guandor: Thanks, fixed.

ports/balatro/port.json
Guandor: You are right, thank you, fixed.

JanTrueno (approved):

Dia2809 (approved):

PR #2612, opened 2026-08-11
Update porter name

mxmgorin:
I’m using a consistent nickname, so I’d like to update it in the ports as well, if possible.


JanTrueno (approved):
I think this should be fine, please also change your details in the porters.json :)

Cebion (approved):

